PR #36, opened 2024-05-03
Move PTAXSIM DB file to runner temp dir

dfsnow:
This PR fixes two bugs related to our PTAXSIM CI integration. Mainly:

- Newer versions of R seem to copy the contents of the package directory to a temporary folder when building a package. Usually this is fine, but GitHub runners only have a limited amount of disk space. The unzipped PTAXSIM DB file was being copied from the working directory to the temp build directory, exhausting the disk space on the runner. To fix this, I moved to the PTAXSIM DB file to the temp directory and pointed all connection strings to the new location.
- One of the tests pulls a random selection of 1M PINs and compared the PTAXSIM-calculated tax bill to the real tax bill. This test was [intermittently failing](https://github.com/ccao-data/ptaxsim/actions/runs/8944108955/job/24570310257). The failure turned out to be due to a single agency/base combo that resulted in a `NaN` tax rate (divide by 0 error). I added a line to replace `NaN` values with 0 and added a test to catch future issues of the same nature.

### DIFF
--- a/vignettes/appeals.Rmd
+++ b/vignettes/appeals.Rmd
@@ -50,6 +50,14 @@ library(sf)
 ptaxsim_db_conn <- DBI::dbConnect(RSQLite::SQLite(), here("./ptaxsim.db"))
 ```
 
+```{r, echo=FALSE}
+# This is needed to build the vignette using GitHub Actions
+ptaxsim_db_conn <- DBI::dbConnect(
+  RSQLite::SQLite(),
+  Sys.getenv("PTAXSIM_DB_PATH")
+)
+```
+
 ## Gathering PINs of interest
 
 To determine the impact of appeals, we first need a way to gather all the properties (PINs) in Schaumburg. Fortunately, PTAXSIM's database has all the data required to accomplish this task.
@@ -228,7 +236,7 @@ sb_pins_summ <- sb_pins_all %>%
     ),
     idx = (med_av / med_av[stage == "2018\nboard"]) * 100
   )
-```  
+```
 
 Finally, we can plot the index over time to see how reassessment and the subsequent appeals at each stage impacted assessed values.
 
--- a/vignettes/exemptions.Rmd
+++ b/vignettes/exemptions.Rmd
@@ -13,7 +13,7 @@ knitr::opts_chunk$set(
 
 # Introduction
 
-Property tax exemptions are savings that lower a homeowner’s property tax bill. They work by reducing or freezing the taxable value (Equalized Assessed Value, or EAV) of a home. For example, the most common exemption, the Homeowner Exemption, reduces EAV by \$10,000. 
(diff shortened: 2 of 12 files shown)


vignettes/exemptions.Rmd
```diff
@@ -13,7 +13,7 @@ knitr::opts_chunk$set(
 
 # Introduction
 
-Property tax exemptions are savings that lower a homeowner’s property tax bill. They work by reducing or freezing the taxable value (Equalized Assessed Value, or EAV) of a home. For example, the most common exemption, the Homeowner Exemption, reduces EAV by \$10,000.
```
dfsnow, line 16: This stuff is just extraneous cleanup from my built-in linter/fixers.

vignettes/mapping.Rmd
```diff
@@ -41,6 +41,14 @@ library(tidyr)
 ptaxsim_db_conn <- DBI::dbConnect(RSQLite::SQLite(), here("./ptaxsim.db"))
 ```
 
+```{r, echo=FALSE}
+# This is needed to build the vignette using GitHub Actions
+ptaxsim_db_conn <- DBI::dbConnect(
+  RSQLite::SQLite(),
+  Sys.getenv("PTAXSIM_DB_PATH")
+)
+```
```
dfsnow, line 50: Vignettes need to point at the temp directory during build time, but I don't necessarily want to use the env var setup here since it's a bit confusing.

vignettes/appeals.Rmd
```diff
@@ -50,6 +50,14 @@ library(sf)
 ptaxsim_db_conn <- DBI::dbConnect(RSQLite::SQLite(), here("./ptaxsim.db"))
 ```
 
+```{r, echo=FALSE}
+# This is needed to build the vignette using GitHub Actions
+ptaxsim_db_conn <- DBI::dbConnect(
+  RSQLite::SQLite(),
+  Sys.getenv("PTAXSIM_DB_PATH")
+)
```
jeancochrane, line 58: **[Question, non-blocking]** What's the relationship between this `DBI::dbConnect()` call and the one that precedes it on line 50? Is the idea just that this will overwrite the preceding `ptaxsim_db_conn` definition on CI, but locally we can skip this call and use the one on line 50 instead?

jeancochrane (approved):
Nice work, looks great! I don't totally get the different circumstances under which the database connection is supposed to be created for the vignettes, but I'm fine moving forward if you feel good about this pattern.

vignettes/appeals.Rmd
dfsnow: You got it. This call is hidden in the actual output vignette HTML just because I think it's simpler to show a file path instead of an env var setup, but the env var is necessary for it to work on CI.

PR #37, opened 2024-05-28
Fix file permissions error in `style-files` pre-commit hook

jeancochrane:
This PR implements the same change as https://github.com/ccao-data/model-res-avm/pull/244 in order to fix the same bug in the `pre-commit` workflow. See that PR for more background on this change.

In addition, we switch over to the https://github.com/ccao-data/actions repo for the `pre-commit` workflow, to bring this repo in line with our other repos that use pre-commit like `model-res-avm`.

I didn't test timing for serial execution vs parallel execution in the same way that I did in https://github.com/ccao-data/model-res-avm/pull/244 and https://github.com/ccao-data/model-condo-avm/pull/49 since the results of those two PRs have me convinced that serial execution is likely better and at least no worse.

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -12,27 +12,5 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Setup python
-        uses: actions/setup-python@v4
-        with:
-          python-version: 3.x
-
-      - name: Install pre-commit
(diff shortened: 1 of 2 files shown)


wrridgeway (approved):

jeancochrane:
Here again I need @dfsnow's review as a codeowner ☹️

wrridgeway (approved):

dfsnow (approved):

PR #42, opened 2024-08-05
Remove codecov integration

dfsnow:
Super quick PR to remove the failing codecov integration.


jeancochrane (approved):
Looks good! Will this alert us in the case of a coverage decrease, or does it just report the status in the workflow logs?

dfsnow:
> Looks good! Will this alert us in the case of a coverage decrease, or does it just report the status in the workflow logs?

Only the latter unfortunately. Let's save the alert/error on decrease for a later issue.

Issue #5, opened 2023-07-10
Create vignette on levy changes, inflation, PTELL, etc


dfsnow:
Need to create a vignette to show how levy changes affect tax bills. Specifically, we should show how inflation relates to levy increases, how PTELL works, what the effect of inflation will be on levies, etc.

erhla:
tbd https://github.com/erhla/ptaxsim/tree/inflation-ptell

PR #47, opened 2024-12-17
Inflation ptell

erhla:
Proposed vignette for #5 

PTELL is a little complicated so open to other approaches. 

A final issue I haven't yet found the best way to raise is that the year over year percent increase in the capped extension exceeds the PTELL limit BUT that doesn't necessarily mean that the year over year increase for an individual taxpayer would since the PTELL limit is designed to let a taxing body capture new property and expired tif increment etc which generally is referred to as not a tax increase for existing taxpayers even though it leads to increased revenue. That being said, I think being able to show this relationship is a really interesting one but the ways I've played around don't communicate the issue clearly. I tried calculating a tax bill with a constant EAV and also simulating a tax bill which started at a specific EAV in year 2006 and then mirrored the percent changes in the total EAV available to CPS every year. Both were attempts to remove the impacts of relative AV changes (the % share that a property has of the total EAV) to isolate the changes in capped/uncapped tax bills.

### DIFF
--- a/vignettes/levies.Rmd
+++ b/vignettes/levies.Rmd
@@ -15,28 +15,43 @@ library(DBI)
 library(tidyverse)
 library(ptaxsim)
 
+theme_set(theme_bw())
 
-make_table <- function(inputdata){
-    DT::datatable(
-      inputdata, extensions = 'Buttons', 
-      rownames = FALSE,
-      options = list(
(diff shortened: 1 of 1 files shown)


erhla:
Working on a larger report here w/ some other groups. Happy to circle back to this whenever.

PR #48, opened 2024-12-17
Fix pkgdown build on GitHub Actions for new Ubuntu 24.04 runners

jeancochrane:
This PR fixes a problem with the new Ubuntu 24.04 runners on GitHub Actions that is [causing pkgdown builds to fail](https://github.com/ccao-data/ptaxsim/actions/runs/12377159382/job/34546010640). See https://github.com/ccao-data/data-architecture/pull/683 for detailed background on the bug and its resolution.


dfsnow (approved):
Thanks 👍

Issue #46, opened 2024-12-12
Appeals vignette agency totals issue


cctosilber:
Hi,

I was trying to use the appeals vignette to look at tax bill changes across the county due to appeals - essentially the same task as the vignette, but for all PINs in Cook County instead of just one municipality. I noticed that in the scenarios where appeals are removed, the countywide tax amount increases substantially for nearly all taxing agencies. I'm not sure if this is a bug, if I'm doing something wrong, if there's a reason that non-TIF extensions would increase without appeals, or if the tool is simply not designed for an analysis like this.

Cook County had a levy of $782 million in 2021, and the extension using the "board" AV was very close to this, at $784 million. The extension amount ballooned to $910 million using "mailed" AV. As I understand it, this shouldn't happen - the amount extended for the county should be roughly the same, and different AV amounts should only shift the share of this levy that individual properties pay.

Anyway, I'm trying to figure why this is happening. If you have any insight, please let me know.

Thanks,
Chris

Here is my script. It's very close to the original vignette, with Cook County's agency number instead of Schaumburg and 2021 instead of 2018. At the end, I group the results by agency to compare the agency totals in different scenarios:

```
library(DBI)
library(data.table)
library(dplyr)
library(ggplot2)
library(ggspatial)
library(glue)
library(here)
library(ptaxsim)
library(sf)

options(scipen=999)

# connect to PTAXSIM

ptaxsim_db_conn <- DBI::dbConnect(RSQLite::SQLite(), "\\ptaxsim-2023.0.0.db\\ptaxsim-2023.0.0.db")

# 2021 ----

# get 2021 PINs by finding all tax codes billed to Cook County

tax_codes21 <- DBI::dbGetQuery(
  ptaxsim_db_conn, "
  SELECT tax_code_num
  FROM tax_code
  WHERE agency_num = '010010000'
  AND year = 2021
  "
)

pins21 <- DBI::dbGetQuery(
  ptaxsim_db_conn,
  glue_sql("
    SELECT pin, class
    FROM pin
    WHERE tax_code_num IN ({tax_codes21$tax_code_num*})
    AND year = 2021
    ",
           .con = ptaxsim_db_conn
  )
)

# Grab equalization factors from the PTAXSIM database

eq_dt <- DBI::dbGetQuery(ptaxsim_db_conn, "SELECT * FROM eq_factor")

# Calculate tax bills from 2021 at each stage of appeal. Note that recalculating
[...]

PR #50, opened 2025-01-29
Fix appeals vignette EAV aggregation method

kyrasturgill:
While reviewing possible reasons for [Appeals vignette agency totals issue #46 ](https://github.com/ccao-data/ptaxsim/issues/46), an error was identified in the way the appeals vignette adds pre-appeal EAV back into the base for taxing districts. Currently the code aggregates the EAV difference between stages at the tax code level, while it should aggregate up to the agency. This results in agencies that cover multiple tax codes to have their agency rates vary across tax codes, while agencies should have a uniform tax rate regardless of the tax code. 

This PR adjusts the code to ensure the EAV difference within each tax code is summed across all tax codes of an agency. This fix does help in aligning the summed amount taxed by agencies between stages, which was the issue identified in issue #46 (this flagged that taxes collected by Cook County in 2021 would be 16% higher in the mailed stage than the board stage, when they both should theoretically remain close to that agency's extension). With this adjustment, the entirety of taxes paid Cook County for the mailed and board stages vary by 3.8% rather than 16%.

### DIFF
--- a/vignettes/appeals.Rmd
+++ b/vignettes/appeals.Rmd
@@ -363,8 +363,10 @@ sb_bills_all <- purrr::map_dfr(c("mailed", "certified", "board"), function(x) {
     left_join(eq_dt_stage, by = "year") %>%
     # Aggregate the total differences for each district
     group_by(agency_num, agency_name) %>%
-    mutate(eav_diff_total = sum(eav_diff_total),
-           exe_diff_total = sum(exe_diff_total)) %>%
+    mutate(
+      eav_diff_total = sum(eav_diff_total),
+      exe_diff_total = sum(exe_diff_total)
+    ) %>%
     ungroup() %>%
     mutate(
       # Recalculate the base in terms of the "tentative" or "final" equalizer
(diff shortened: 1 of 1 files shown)


kyrasturgill:
This change in methodology results in greater differences between median residential tax bills - median residential tax bill in 2019 certified stage increases by 11% rather 7%. This has been adjusted in the text of the vignette. 
Perc diff between residential median 2019 certified and 2019 board bills:
 `(5068.47-4565.77)/4565.77 = .11`

Other figures include the percent differences between median commercial bills at different stages:
Perc diff between commercial median 2019 mailed and 2018 board:
`(70219.01-55332.65)/55332.65 = .27`

Perc diff between commercial median 2019 mailed and 2019 board:
`(57906.71-70219.01)/70219.01 = -.17`

dfsnow (approved):
@kyrasturgill Nice work here, aggregating only to the tax code level was definitely wrong. This doesn't totally fix #46, but it moves things in the right direction. Let's merge it but keep #46 open.

PR #52, opened 2025-03-07
Fix CI builds

jeancochrane:
This PR fixes two problems with our CI builds:

* `actions/cache@v3` [is deprecated](https://github.com/orgs/community/discussions/142581) and will throw an error unless updated to v4.
* The `lint` workflow and the `pre-commit` workflow both run lintr, but they have a version mismatch due to [new default linters in v3.2.0](https://github.com/r-lib/lintr/releases/tag/v3.2.0). In short, 3.2.0 makes `return_linter` the default, which is a style that we generally don't follow; however, we can't disable it in the `.lintr` config file because pre-commit is not on version 3.2.0 yet and so does not consider `return_linter` to be the default. Rather than do surgery to resolve the version mismatch, I decided to remove the `lint` workflow, which is unnecessary given that we're already running `lintr` in the `pre-commit` workflow anyway.


wrridgeway (approved):
Thanks Jean!

Issue #56, opened 2025-07-28
Transit TIFs


erhla:
Everyone's favorite type of TIFs (Transit TIFS, RPM and RLE). Will keep updating/debugging this.

- [x] RLE tif. Need to add.
- [x] For `simplify=TRUE`, should we report the share the TIF pays to CPS as part of the TIF line item (as current) or just the share to the TIF and add the TIF share to CPS as a separate item (how the Treasurer does it on bills) or combined item w/ the CPS line item
- [x] Clarify difference between how clerk reports total TIF revenue by tax code and how PTAXSIM does
- [x] The Cook County line item is not the same between actual tax bills and the expected formula (20% of non CPS portion of TIF revenue)

jeancochrane:
@erhla The missing RLE TIF and the `simplify=TRUE` changes make sense to me and @kyrasturgill. I think that we both understand the other two issues at play here, but let me repeat them back to you in our own words to make sure we've got it:

1. The revenue columns in the Clerk's [TIF Agency Distribution Summary reports](https://www.cookcountyclerkil.gov/property-taxes/tifs-tax-increment-financing/tif-reports) (`Tax Code Revenue` and `TIF Total Revenue`) do not match the sum of transit TIF revenues across all tax bills for tax codes in PTAXSIM because the Clerk does not remove agency distributions from those columns, while PTAXSIM _does_ remove agency distributions from its tax bills.
    * More specifically, are we correct that the mismatch here is between the total revenue columns in the Clerk's report on the one hand, and _the sum of transit TIF revenues across all individual tax bills in a tax code_ in PTAXSIM on the other?
2. The transit TIF line items in the taxpayer-facing tax bills always show $0 for the transit TIFs and the agencies that should receive distributions from those TIFs, even if the bill itself is paying into a transit TIF.
    * If this is correct, then how are the taxpayer-facing tax bills incorporating these amounts that are otherwise missing?

Also, just to be extra clear, we want to make sure we understand what you're saying here:

> The reason that the tax code revenue from the TIF agency reports is not the same as ptaxsim/tax_bill is because the variable final_tax_to_tif does not include the 20% distribution to other taxing bodies but does include the 53% payment to CPS. This is confusing. The clerk includes both the payment to CPS and the payments to other taxing districts in their number.

[...]

PR #58, opened 2025-08-01
Add RLE transit TIF and fix `simplify = TRUE` logic to account for transit TIFs

erhla:
RE #56. I'd propose squashing this in if accepted since it took me a minute to figure out a few issues. 

This:

- Adds in the RLE tif into the transit tif calculations. (It was previously missing)
- Spins off the transit tif calculations into a nested function in tax bill. (not strictly necessary but it is more clear)
- Proposes adding a new version of `simplify=TRUE` for the `tax_bill` function called `simplify_transit`. I created this as an additional parameter but I would recommend that `simplify=TRUE` be replaced with `simplify_transit` since `simplify` is designed to mirror the agency line items on the tax bill but does not show agency line items on the tax bills in transit tifs as they actually are. AFAIK there shouldn't be any issues with replacing simplify with simplify_transit but I wanted to defer to you instead of proposing that directly. Let me know and I can make that change.
- Rename all the `RPM_` specific transit TIF variables to be named `transit_` since there are now two transit TIFs.
- update `final_tax_to_tif` to include `final_tax_to_dist`. The Clerk TIF revenue by tax code report includes the TIF revenue before either the CPS or non-CPS distributions are made from Transit TIFs. This means that if you sum the `final_tax_to_tif` it will align with the Clerk report.

This does not:

- Fix the issue identified related to how the actual tax bills calculate the agency line items which appear to be incorrect. The PTAXSIM calculations are technically correct but it appears that the Clerk/Treasurer are doing something a little different.


jeancochrane:
Thanks for this @erhla! I'm going to take a look early next week. These days @kyrasturgill gets final say over this repo, and she's out on vacation currently, so we won't be able to merge till the week of 8/11 at the earliest.

erhla:
I talked with the Treasurer today. They confirmed that the line items on the transit tax bills do not reflect the actual distributions to the agencies and that we should not try to replicate those calculations. They will send me the actual distribution percentages by tax code so I will verify that our calculations are correct compared to that.

jeancochrane (changes requested):
This looks awesome, thanks @erhla! Two chores that @kyrasturgill and I think you should do before we merge:

1. Consolidate `simplify` and `simplify_transit`
    * If we're understanding your suggestion correctly, then we agree with you that we should simply replace the `simplify` logic with `simplify_transit`, but continue using the parameter name `simplify`; if that's a misunderstanding of your suggestion, let us know and we can talk it through
3. Tweak unit tests to make sure they pass
    * Neither of us have inspected the unit tests in great detail yet, but it may make sense to add a test specific to the transit TIF simplification logic, since it's slightly different from the other simplification logic; we're open to that not being necessary if you think the existing `simplify` tests cover the new logic well enough, though

Once those changes are in, we'll do some more rigorous QC and then approve once all looks good.

erhla:
Now updated `final_tax_to_tif` to include all three components: example for 2023

image

jeancochrane:
@erhla I fixed a bug in #60 that I believe has been causing all of your CI checks to fail. Once you're ready to work on getting CI checks to pass, try merging in (or rebasing off of) the latest commit to the main branch of `ccao-data/ptaxsim` to fix that bug on your branch.

erhla:
Alright I think all looks good here. I added a few tests to ensure that the transit tif / simplify changes calculate things correctly. I was thinking about adding a vignette but I feel like transit tifs are not worth their own vignette and also adding them into the tinkering with tifs vignette feels kind of random.

R/tax_bill.R
```diff
@@ -58,10 +58,16 @@
...
 #'   and \code{tax_code_vec}.
-#' @param simplify Default \code{TRUE}. Boolean to keep only the columns that
-#'   appear on a real tax bill. Additionally, collapses the TIF output column
-#'   \code{final_tax_to_tif} to a line-item, similar to the format on a real
-#'   tax bill.
+#' @param simplify Default \code{TRUE}. Boolean to keep only the columns
+#'   that appear on a real tax bill
+#'   which also includes the "Board of Education - from Transit TIF" line.
+#'   Additionally, collapses the TIF output column \code{final_tax_to_tif}
+#'   to a line-item, similar to the format on a real tax bill. Note: In order to
```
jeancochrane, line 65: **[Nitpick, optional]** It took me a while to parse what these docs were trying to say, so here's an attempt at an improvement that I think will be clearer to someone who is using this package for the first time:

```suggestion
#' @param simplify Default \code{TRUE}. Boolean to keep only the columns
#'   that appear on a real tax bill.
#'   If the parcel is in a TIF, simplified output will collapse the TIF output
#'   column \code{final_tax_to_tif} to a line item, similar to the format on a
#'   real tax bill; further, if the parcel is in a transit TIF, simplified
#'   output will also collapse the column \code{transit_tif_to_cps} to a line
#'   item called "Board of Education - from Transit TIF". Note: In order to
#'   more completely show information for transit TIF districts both
#'   \code{final_tax_to_tif} and \code{final_tax_to_dist} include
#'   \code{transit_tif_to_dist}. Therefore, to calculate final total taxes
#'   when \code{FALSE}, use
#'   \code{final_tax_to_tif + final_tax_to_dist - transit_tif_to_dist}.
```

tests/testthat/test-tax_bill.R
```diff
@@ -145,16 +147,28 @@ test_that("returned amount/output correct for all sample bills", {
...
       nrow(),
-    775
+    781
   )
 
+  base_bills <- tax_bill(sum_dt$year, sum_dt$pin, simplify = TRUE) %>%
+    select(year, pin, agency_num, final_tax)
+  transit_bills <- base_bills %>%
+    count(pin, agency_num) %>%
+    filter(n > 1)
```
jeancochrane, line 157: **[Nitpick, optional]** It took me a minute to figure out why this filter would return only transit bills, so a comment for posterity:

```suggestion
    # PINs in transit TIFs will have two line items with the CPS agency number,
	# one being the TIF distribution
    filter(n > 1)
```

tests/testthat/test-tax_bill.R
```diff
@@ -145,16 +147,28 @@ test_that("returned amount/output correct for all sample bills", {
...
+
   # District level tax amounts
   expect_equivalent(
     tax_bill(sum_dt$year, sum_dt$pin, simplify = TRUE) %>%
       select(year, pin, agency_num, final_tax) %>%
+      filter(!pin %in% transit_bills$pin) %>%
+      group_by(year, pin, agency_num) %>%
+      summarize(final_tax = sum(final_tax)) %>%
+      # combine cps transit tif rows w/ tif row
```
jeancochrane, line 166: **[Nitpick, optional]** I think moving this comment to the top of the grouping block will make it clearer:

```suggestion
      # combine cps transit tif rows w/ tif row
      group_by(year, pin, agency_num) %>%
      summarize(final_tax = sum(final_tax)) %>%
```

But also, do we even need to do this, given that we're filtering for `!pin %in% transit_bills$pin` above?

jeancochrane (approved):
Really nice work here! No serious concerns from my end, just a bunch of clarifying questions and small nitpicks that I think will make the code easier to read for other rookies like me. We're at the nitpick stage, so I'm happy to do the legwork of resolving all of our remaining nits and getting the PR merged if you'd prefer that.

erhla:
Thanks! I think all the comments both of you made are clear to me.

I've enjoyed this. Normally I'm just a team of one so this was really fun for me, but please squash all my commits.

Now we just have to monitor how the Treasurer is going to update this for the 2024 bills.

jeancochrane:
@erhla It seems like you resolved a few of my questions without adding answers, I wonder if you added comments but then didn't submit the review?

> I've enjoyed this. Normally I'm just a team of one so this was really fun for me, but please squash all my commits.

Agreed, it's also a pleasure for us to work with someone externally who has the technical chops to make a contribution like this!

erhla:
>  It seems like you resolved a few of my questions without adding answers, I wonder if you added comments but then didn't submit the review?

Ah for those I accepted your suggestion as is and then resolved your question. I haven't used github like this before so let me know if there's something different I should have done.

R/tax_bill.R
```diff
@@ -58,10 +58,23 @@
...
+#'   item called "Board of Education - from Transit TIF". Note: In order to
+#'   more completely show information for transit TIF districts both
+#'   \code{final_tax_to_tif} and \code{final_tax_to_dist} include
+#'   \code{transit_tif_to_dist}. Therefore, to calculate final total taxes
+#'   when \code{FALSE}, use
+#'   \code{final_tax_to_tif + final_tax_to_dist - transit_tif_to_dist}.
+#'   more completely show information for transit TIF districts both
+#'   \code{final_tax_to_tif} and \code{final_tax_to_dist} include
+#'   \code{transit_tif_to_dist}. Therefore, to calculate final total taxes
+#'   when \code{FALSE}, use
```
jeancochrane, line 76: **[Nitpick, required]** Looks like we accidentally duplicated some lines of this docstring:

```suggestion
```

We'll also need to regenerate the autogenerated `.Rd` docs after we make this change.

erhla (commented):
Turns out I didn't realize you have to commit the comments! So here's my answers (belatedly).

erhla:
I'm looking at the new tax bills for tax year 2024 and the treasurer has changed so many things tragically. I'm hoping they might clean it up more but initially it looks like it is much less clear. I think we should ideally merge this in for ty23 and it probably makes sense to show the information in the most clear way we believe for ty24 and not how the treasurer is doing it. I want to see some of the tif reports from the clerk first to see if the calculations are even being doing correctly.

erhla:
image

Here is an example. Note that the Board of Education TIF and regular levy is now combined into one item confusingly labeled board of education TIF RPM1

R/tax_bill.R
```diff
@@ -299,16 +351,19 @@ tax_bill <- function(year_vec,
         "agency_tax_rate", "final_tax"
       )
     )
-  } else {
+  }
+
+  if (!simplify) {
```
jeancochrane: **[Nitpick, required]** For the sake of cleanliness, let's revert to `else` here, since it's a more standard way to mark the false branch of a conditional expression:


```suggestion
  } else {
```

jeancochrane (approved):
Thanks for your patience @erhla! Pending your resolution of one final nitpick from my end, this is good to merge. I'll look out for a commit from you on that front and merge once it lands.

Heads up that I tweaked the title of the PR to be more descriptive -- feel free to change it if you think I've misrepresented the PR at all.

> I'm looking at the new tax bills for tax year 2024 and the treasurer has changed so many things tragically. I'm hoping they might clean it up more but initially it looks like it is much less clear. I think we should ideally merge this in for ty23 and it probably makes sense to show the information in the most clear way we believe for ty24 and not how the treasurer is doing it. I want to see some of the tif reports from the clerk first to see if the calculations are even being doing correctly.

Agreed. I added a quick note about this to our [2024 refresh omnibus issue](https://github.com/ccao-data/ptaxsim/issues/59) to make sure we don't lose track of it. In the meantime, I think the diff in this PR is a big improvement.

erhla:
Amazing, thank you! re #59, based on what the Treasurer said it looks like TIF calculations are done by pin (and the pin's frozen eav at the time the TIF was created) now instead of tax code which is going to be interesting to say the least. So I think we will permanently need to calculate TY23 and earlier tax bills and TY24 and later bills with different code.

jeancochrane:
> based on what the Treasurer said it looks like TIF calculations are done by pin (and the pin's frozen eav at the time the TIF was created) now instead of tax code which is going to be interesting to say the least. So I think we will permanently need to calculate TY23 and earlier tax bills and TY24 and later bills with different code.

Agreed, I just updated the issue text to reflect this.

Thanks again for your hard work getting this landed!

PR #62, opened 2025-10-21
Replace hyperlink to IDOR assessment ratios with 2023 report

kyrasturgill:
This PR is to replace the hyperlink in the tax rates vignette that refers to the IDOR assessment ratios that were used for deriving the tax base increase. The vignette was linking to IDOR's 2022 Assessment Ratio report, while it really should be linking to the 2023 Assessment report as our methodology uses 2023 ratios. All references to ratios are correct, so no changes are required there - it is just the hyperlink that needed to be adjusted.


jeancochrane (approved):
Nice and simple!

PR #63, opened 2026-01-02
Add 2024 AVs, exemptions, equalizer, and CPI

jeancochrane:
This PR tweaks a few `data-raw` scripts to add 2024 data to the `pin`, `cpi`, and `eq_factor` tables. I have already used this code to load the corresponding files into the testing bucket on S3.

The most complicated of these changes relates to the `pin` table, whose data source needs to change in 2024 following the Clerk's migration from the AS400 to iasWorld as their source-of-truth database. Rather than pull AV and exemption data from a SQL server mirror of the AS400, as we used to do, we now pull these data from a flat file stored in S3. In future years, we may pull this data from iasWorld directly, so I did a little bit of QC work to check the flat file against iasWorld; they mostly match up, though there remain a few thousand rows with discrepancies that I couldn't track down. (See EI issue  395, which will investigate these discrepancies in more detail.)

Note that this PR doesn't yet include changes to import PIN geometry, because I haven't been able to get that to run successfully yet.

Connects #59.


data-raw/cpi/cpi.R
```diff
@@ -27,6 +27,13 @@ cpi <- pdftools::pdf_text(pdf = "data-raw/cpi/cpihistory.pdf") %>%
   unlist() %>%
   tibble(vals = `.`) %>%
   mutate(vals = str_squish(vals)) %>%
+  # Remove footer lines that do not contain any data
+  filter(
+    !str_detect(
+      vals,
+      regex("printed by the authority|ptax-115", ignore_case = TRUE)
+    )
+  ) %>%
```
jeancochrane, line 36: This footer appears to be new as of 2025. See here to examine it: https://tax.illinois.gov/content/dam/soi/en/web/tax/localgovernments/property/documents/cpihistory.pdf

data-raw/pin/pin.R
```diff
@@ -1,110 +1,300 @@
...
 remote_path_pin_geometry_raw <- file.path(
   remote_bucket,
   "pin_geometry_raw",
   "part-0.parquet"
 )
 
+# Start and end years of data to query, inclusive.
+# Set these to the same value if you want to update only one year of data
+start_year <- 2006
+end_year <- 2024
```
jeancochrane, line 28: Adding these in so that we have an easy way of skipping prior years of data whenever we do an update. This is particularly useful right now because I don't have access to the AS400 mirror, which is required in order to reproduce pre-2024 data. At some point we should get that set up, but I don't want it to block us right now.

data-raw/pin/pin.R
```diff
@@ -1,110 +1,300 @@
...
-# values come from the legacy CCAO database, which mirrors the county mainframe
-ccaodata <- dbConnect(
-  odbc::odbc(),
-  .connection_string = Sys.getenv("DB_CONFIG_CCAODATA")
-)
+# Get data frame of all AVs, tax codes, and exemptions per PIN from 2006 to
+# 2023. These values come from the legacy CCAO database, which mirrors the
+# county mainframe.
+# Only query this data if we are pulling data for years up to 2023
+if (start_year <= 2023) {
```
jeancochrane, line 37: There are a few conditional branches in this file that split depending on whether we're ingesting data before or after 2023. I considered creating a new file dedicated exclusively to post-2024 data manipulation, since it feels to me like this file will get very messy very fast if they substantially change the data model again in the future (causing us to need to introduce further conditional branches based on year). For now, however, modifying this file feels like the simpler path, and I expect it's also easier to review that way.

data-raw/pin/pin.R
```diff
@@ -1,110 +1,300 @@
...
+      ORDER BY C.TAX_YEAR, C.PIN, C.CL_ASSD_VAL DESC
+    ", .con = ccaodata)
+  ) %>%
+    mutate(
+      across(c(year, pin, tax_code_num, class), as.character),
+      across(c(starts_with("av_"), starts_with("exe_")), as.integer),
+      exe_homeowner = ifelse(exe_homeowner < 0L, 0L, exe_homeowner),
+      tax_bill_total = tidyr::replace_na(tax_bill_total, 0),
+      # This exemption is new in 2024 and does not exist in the legacy data
+      exe_vet_dis_100 = 0L
```
jeancochrane, line 99: This is the second change I made to the pre-2023 query.

data-raw/pin/pin.R
```diff
@@ -1,110 +1,300 @@
...
+    AS INT) AS exe_vet_dis_ge70,
+    CAST(
+      SUM(
+        CASE WHEN exe_name = 'exe_vet_dis_100' THEN exe_amount ELSE 0 END
+      )
+    AS INT) AS exe_vet_dis_100
+  FROM long
+  GROUP BY pin, year
+  ", .con = ccaoathena)
+) %>%
```
jeancochrane: Much of this query logic duplicates the logic in [`default.vw_pin_exe`](https://github.com/ccao-data/data-architecture/blob/master/dbt/models/default/default.vw_pin_exe.sql). However, that view does not provide a method for filtering by CofE (https://github.com/ccao-data/data-architecture/issues/962), which is important in this context. As such, I've decided to duplicate the logic for now, and we can clean this up later once we incorporate CofE flags into the data lake.

data-raw/pin/pin.R
```diff
@@ -147,16 +416,31 @@ agency_df <- read_parquet(file.path(remote_bucket, "agency", "part-0.parquet"))
...
 # Grab parcel shapes from S3. These files are originally from the data portal
-# and Cook Central
-remote_bucket_geometry <- "s3://ccao-data-warehouse-us-east-1/spatial/parcel"
-pin_geometry_df_full <- arrow::open_dataset(remote_bucket_geometry) %>%
-  filter(year >= 2006 & year <= max_year) %>%
-  select(year, x = lon, y = lat, pin10, geometry) %>%
-  geoarrow_collect_sf()
+# and Cook Central. We have to load each parquet file individually instead of
+# loading them all as a Dataset because some issue with the file metadata causes
+# an esoteric error when geoarrow tries to collect the files as a Dataset
```
jeancochrane: Neither Billy nor I are sure why this just started breaking recently. However, it prevents us from using `geoarrow_collect_sf()`, so I refactored to load each Parquet file individually instead -- an ugly solution, but it works.

data-raw/pin/pin.R
```diff
@@ -1,110 +1,262 @@
...
+          ON C.PIN = TB.PIN
+          AND C.TAX_YEAR = TB.TAX_YEAR
+      LEFT JOIN AS_HEADBR BR
+          ON C.PIN = BR.PIN
+          AND C.TAX_YEAR = BR.TAX_YEAR
+      LEFT JOIN TAXBILLAMOUNTS BILLS
+          ON C.PIN = BILLS.PIN
+          AND C.TAX_YEAR = BILLS.TAX_YEAR
+      WHERE C.TAX_YEAR >= {start_year}
+        AND C.TAX_YEAR <= 2023
```
jeancochrane, line 89: This is one of two changes I made to the pre-2023 query.

data-raw/pin/pin.R
```diff
@@ -128,9 +280,88 @@ pin_fill <- pin %>%
...
+    exe_vet_dis_100 = ifelse(
+      exe_vet_dis == exe_vet_dis_tot_athena,
+      exe_vet_dis_100_athena,
+      ifelse(
+        exe_vet_dis > 0 & exe_vet_dis_100_athena > 0,
+        exe_vet_dis,
+        0L
+      )
+    )
+  ) %>%
```
jeancochrane, line 356: The logic here is pretty complicated, but I think it's important that we get it right, since we're making some interpretive decisions here rather than pulling information directly from the tax bill export. Let me know if it would help to walk through it together.

data-raw/pin/pin.R
kyrasturgill: Got it, so if I'm understanding correctly, to run this just for 2024 update we'd define `start_year` and `end_year` as 2024?

data-raw/pin/pin.R
kyrasturgill: I'm pretty sure I've got it! Did you have the chance to test the mismatch PINs to confirm the provided exemption amounts, EAV and tax rate result in the tax bill total provided?
It makes sense to me to prioritize accuracy of the tax bill export especially if using the athena exemption amounts led to a calculated tax bill amount that does not match `tax_bill_total`. 
And then the logic of keeping the exemption amount in the same exemption tier seems the most straightforward. I wondered if we could create a rule based on the exemption amount, but those seem to be pretty inconsistent.

kyrasturgill (commented):
This looks great! Thanks for thinking through the veteran's disability logic - I had a couple questions/thoughts but don't have any concrete changes.

data-raw/pin/pin.R
jeancochrane: Yup, that's right!

data-raw/pin/pin.R
jeancochrane: > Did you have the chance to test the mismatch PINs to confirm the provided exemption amounts, EAV and tax rate result in the tax bill total provided?

I did! I don't have the exact results now, but I'll include them in https://github.com/ccao-data/enterprise-intelligence/issues/395. My memory is that there were a couple thousand mismatches whose discrepancy I couldn't nail down; most of them were missing from the Treasurer's tax bill portal. I'll make sure to include that analysis in https://github.com/ccao-data/enterprise-intelligence/issues/395.

PR #65, opened 2026-01-15
Add 2024 agency, tax code and TIF reports

kyrasturgill:
This PR adds the raw data for 2024 tax code rate, agency rate and TIF reports. It also makes necessary adjustments to the data-raw ingest scripts that are necessary on account of changes made to 2024 data structure. 

Most changes to the ingest scripts involve slight adjustments to the code for renaming fields.  This code was originally written to work assuming the same fields being selected from the report files exist across all years (which are contained in separate files). The introduction of new fields in the 2024 files led to errors within the `rename_with()` function when it was returning a vector with length of 1 rather than 0, even when the field did not exist. The work around for this was using `rep()` with `rename_with()`: `rename_with(~rep("agency_name", length(.x)), any_of(c("authority_name")))` which would return empty vector if there were no fields present named `authority_name`. 

Other changes were removing certain fields no longer present in 2024 reports, and the creation of the field `fund_type_num` to account for 2024 changes to fund number structure which are now 6 digits rather than 3. `fund_type_num` is the first 3 digits of `fund_num`, which should be consistent across all years. `fund_num` pre-2024 is also now padded with trailing zeros. Because 2024 is now reporting funds at a more detailed level than in prior years, any time trend analysis of funds should use `fund_type_num`.

Something I have not added to the data yet is `agency_num_legacy` or `authority_num`. I realized that we would need to alter prior years' `agency_num` and `agency_name` to align with the revised 2024 `agency_num`. To avoid altering source data, we could simply have an agency crosswalk added to the db that connects the new agency number, the legacy number and authority number, which would be available for user if they did want to do an analysis of agency extensions or agency rates over time. 

Lastly, this PR also brings in a new TIF data source - the `pin_tif_distribution` which is derived from the Clerk's TIF PIN list report.


jeancochrane:
Thanks for this! Very helpful work. A Few clarifying questions:

1) It seems that our general preference is to try to keep the schema of our data tables as close as possible to the schema of the underlying data that we pull from the Clerk. However, we're deviating from that pattern in this PR by adding `agency.fund_type_num` and `pin_tif_distribution.in_transit_tif`. Do we really want to add these fields at the ingest stage? Or do we instead want to derive them on the fly and return them in the `lookup_*()` functions, to keep them separate from the underlying data model? I don't have super strong opinions here, and ultimately I trust you as a frequent PTAXSIM user to decide which of these approaches would be most convenient for the user.

2) I'm uncertain about this comment:

> Something I have not added to the data yet is `agency_num_legacy` or `authority_num`. I realized that we would need to alter prior years' `agency_num` and `agency_name` to align with the revised 2024 `agency_num`.

I thought that `agency_num_legacy` mapped cleanly to prior years' `agency_num`, in which case we could just fill `agency_num_legacy` for pre-2024 years with the value of `agency_num` in that year. I also thought `authority_num` didn't really matter because it's redundant with `agency_num` for 2024, so we might as well bring it in now. But am I misunderstanding? These fields are kinda confusing, so I'm happy to jump on a call if it would be faster to explain.

kyrasturgill:
Here are my thoughts on your questions:
1. I see your point and am definitely good to omit the `in_transit_tif` field because that's not really that helpful to have in the table and we can build that into the `lookup_pin()` function. I do think that `fund_type_num` could be helpful for anyone who has use case for analyzing fund levies over time. `lookup_agency()` function doesn't return fund-level data, so we wouldn't be able to build that new field into the function output, which is why I was thinking it should be somewhere in the db. Maybe it could just live in the `agency_fund_info` table - it seems `capped_fund_ind` is a helper field we derive ourselves so I don't think it would totally break with past practice to include our own derived field there.
2. For `legacy_agency_num` and `authority_num`, I'm very much confusing myself and need to talk it through. For now I'm going to add those to fields as we have them in the drafted proposed schema, and then we can discuss if we want to make changes.

jeancochrane:
> Maybe it could just live in the `agency_fund_info` table - it seems `capped_fund_ind` is a helper field we derive ourselves so I don't think it would totally break with past practice to include our own derived field there.

That makes sense to me! If I understand the data model correctly , it strikes me that `agency_fund_info` is fundamentally a derived table, in that it represents our interpretation of the discrete entities that are present in the raw `agency_fund` table. From that perspective, including further derived information in `agency_fund_info` seems reasonable.

kyrasturgill:
@jeancochrane, this is ready for your review!
Changes include:

- I added the fields to the `agency_info` table that tie the now rolled up agencies to their current "parent" agency. Let me know if you want to talk through my thinking here.
- Additionally I removed the separate "overlap" county fields, summing and replacing them with a consistent `cty_overlap_eav` field in order to have pre-2024 years align with the 2024 structure. 
- Previously PTAXSIM renamed the field `cty_overall_eav` to `cty_total_eav` and removed the original `cty_total_eav` field. There seem to be discrepancies between these two fields only for agencies that fund particular bonds. My guess is there is some special rule for these bonds limiting the tax base to only a portion of the parent taxing agency. In 2024 agency report, the field OverallEAV is no longer present, and when reviewing past tax rate reports I'm more confident that the TotalEAV field reflects the EAV available to this specific agency/fund whereas OverallEAV represented taxable EAV within the entire jurisdiction. The proposed code now used `cty_total_eav` for all years. (_Note of something I haven't yet addressed - it seems that this field didn't exist or was blank in the 2013 report, so currently the value is 0 for `cty_total_eav` in 2013, which we should probably replace with the `cty_overall_eav` just for that year?_)

data-raw/agency/agency.R
```diff
@@ -218,66 +216,76 @@ agency <- map_dfr(file_names, function(file) {
...
+      rowSums(across(starts_with(c(
+        "cty_dupage_eav",
+        "cty_lake_eav",
+        "cty_will_eav",
+        "cty_kane_eav",
+        "cty_mchenry_eav",
+        "cty_dekalb_eav",
+        "cty_kankakee_eav",
+        "cty_grundy_eav",
+        "cty_la_salle_eav",
```
jeancochrane: **[Nitpick, required]** Maybe I'm misreading, but don't we standardize on the name `cty_lasalle_eav` on line 226 above, with no space?

data-raw/agency/agency.R
```diff
@@ -549,6 +554,53 @@ agency_info <- agency_info %>%
...
+
+agency_info <- agency_info %>%
+  left_join(agency_legacy_cw, by = "agency_num") %>%
+  mutate(
+    agency_change_24 =
+      ifelse(
+        agency_num != agency_num_24_update,
+        TRUE,
+        FALSE
+      ),
```
jeancochrane: **[Nitpick, optional]** Maybe it doesn't matter for downstream consumers of this field, but I think this `ifelse()` condition will evaluate to `NA` if either `agency_num` or `agency_num_24_update` are null, which seems plausible given the left join:

```R
> ifelse(NA == TRUE, TRUE, FALSE)
[1] NA
```

If we want to ensure the output is always a boolean, we could use `dplyr::coalesce()` instead:

```suggestion
    agency_change_24 = coalesce(agency_num != agency_num_24_update, FALSE)
```

data-raw/agency/agency.R
```diff
@@ -549,6 +554,53 @@ agency_info <- agency_info %>%
...
+agency_legacy_cw <-
+  openxlsx::read.xlsx(
+    "data-raw/tax_code/2024-tax-code-agency-rate-file.xlsx"
+  ) %>%
+  set_names(snakecase::to_snake_case(names(.))) %>%
+  select(
+    agency_num_24_update = agency,
+    agency_num = legacy_num,
+    authority_num = authority,
+    agency_name_24_update = authority_name
```
jeancochrane: **[Suggestion, optional]** This is purely a style preference, but I wonder if we need the `_update` suffix at the end of these variables, or if the `_24` suffix alone is enough to mark the variable as recording a post-2024 value? The fact that we use `agency_change_24` in the `agency_info` definition below makes me think that perhaps `_24` is good enough as a suffix.

data-raw/agency/agency.R
```diff
@@ -549,6 +554,53 @@ agency_info <- agency_info %>%
...
+        NA
+      )
+  ) %>%
+  select(
+    agency_num,
+    agency_num_24_update,
+    agency_name,
+    agency_name_24_update,
+    agency_name_short,
+    agency_name_original,
```
jeancochrane, line 611: **[Question, non-blocking]** Do you think it's a problem that we don't have short names for agencies that changed in 2024?

data-raw/agency/agency.R
```diff
@@ -549,6 +554,53 @@ agency_info <- agency_info %>%
...
+    major_type,
+    minor_type,
+    agency_change_24
```
jeancochrane, line 616: **[Nitpick, optional]** This is a personal preference and so not necessary to change at all, but since the order of this `select()` call is going to determine the default order of the columns in the output parquet file (and hence the output table), I would advocate that we put the `_24` columns at the end of the table since they are irrelevant to 96% of the rows in the table:

```suggestion
    agency_num,
    agency_name,
    agency_name_short,
    agency_name_original,
    major_type,
    minor_type,
    agency_num_24_update,
    agency_name_24_update,
    agency_change_24
```

data-raw/agency/agency.R
```diff
@@ -218,66 +216,76 @@ agency <- map_dfr(file_names, function(file) {
...
+    )
 }) %>%
   mutate(
+    # One row is missing a Cook EAV value. Fill manually from prior year
+    cty_cook_eav = ifelse(
+      agency_num == "030580002" & year == "2006",
+      0,
+      cty_cook_eav
+    ),
+    across(starts_with("cty_"), ~ replace_na(.x, 0)),
```
jeancochrane, line 267: **[Question, non-blocking]** No action necessary right now, but I'm wondering about this choice to coerce these nulls to 0 in the context of 2024 data, where these County fields are always null. If we were to keep them null, we would hew closer to the actual contents of the input data, where these fields are totally missing. However, that would have the downside of requiring users to handle nulls whenever using these columns. I don't have a good enough grasp of the context to be able to make a decision, but I wanted to raise it as a quirk of the data that I noticed while QCing.

data-raw/agency/agency.R
```diff
@@ -307,11 +310,13 @@ agency <- map_dfr(file_names, function(file) {
...
       c(
         lim_rate, pct_burden, total_prelim_rate, total_final_rate,
-        reduction_pct, total_non_cap_ext, total_ext
+        reduction_pct, total_ext
       ),
       ~ as.double(.x)
     )
-  )
+  ) %>%
+  select(-total_reduced_levy) %>%
```
jeancochrane, line 323: **[Question, non-blocking]** Sorry if I missed the answer to this question in your prior communications, but why remove `total_reduced_levy` here, when we're just going to add it back in after we join to `agency_2013`, which contains the column due to its presence in `agency_fund_2013`?

jeancochrane (commented):
This is getting close to done! I still need to QC the changes to `tif.R`, but I want to send this feedback over to you first so you can get a head start on it.

data-raw/agency/agency.R
kyrasturgill: Nope you read that right and this is a great catch!

data-raw/agency/agency.R
kyrasturgill: Agreed!

data-raw/agency/agency.R
kyrasturgill: Hmmm I don't think I understand, do you mean that we're missing field along lines of `agency_name_short_24`?

data-raw/agency/agency.R
kyrasturgill: Agreed!

data-raw/agency/agency.R
kyrasturgill: I agree - this was bothering me because it is not technically correct to call those 0. I think my preference would be to not replace those NAs with 0, but happy to discuss more.

data-raw/agency/agency.R
kyrasturgill: This was an oversight - I'll remove `total_reduced_levy` it from `agency_2013`.

data-raw/agency/agency.R
```diff
@@ -342,6 +343,7 @@ agency_2013 <- agency %>%
 agency <- agency %>%
   filter(year != 2013) %>%
   bind_rows(agency_2013) %>%
+  select(-cty_overall_eav) %>%
   arrange(year, agency_num)
```
kyrasturgill, line 352: @jeancochrane - I added this code to address `cty_overall_eav` - this field is no longer included in 2024 reports but was previously used to define `cty_total_eav` which is a field in the source data but for some reason it was chosen to replace it with `cty_overall_eav`. I believe `cty_total_eav` from the source data is the correct field to use. The only issue is that the 2013 does not have the `cty_total_eav`, but it does have `cty_overall_eav`. Since we already have this code for account for the oddities of 2013, I tweaked it to also carry over `cty_overall_eav` to replace `cty_total_eav`. As mentioned above, there are a few agencies where there's a slight discrepancy between the two values - so this isn't perfect but I believe a suitable workaround.

data-raw/agency/agency.R
```diff
@@ -579,9 +579,8 @@ agency_legacy_cw <-
 
 # Correct error in Clerk's report which lists incorrect agency number for
 # the TIF VIL OF OLYMPIA FIELDS-GOV HWY/VOLL
-agency_legacy_cw$agency_num[agency_legacy_cw$agency_name == "TIF VIL OF OLYMPIA FIELDS-GOV HWY/VOLL"] <- "030930502"
-agency_legacy_cw$agency_num_24[agency_legacy_cw$agency_name == "TIF VIL OF OLYMPIA FIELDS-GOV HWY/VOLL"] <- "030930502"
-
+idx <- agency_legacy_cw$agency_name == "TIF VIL OF OLYMPIA FIELDS-GOV HWY/VOLL"
+agency_legacy_cw[idx, c("agency_num", "agency_num_24")] <- "030930502"
```
kyrasturgill: The clerk's report erroneously gives this TIF the wrong agency number - the same as TIF VIL OF OLYMPIA FIELDS-GOV HWY/VOLL II - which is why there were duplicate rows generated in `agency_info`.

jeancochrane: **[Suggestion, optional]** Good catch! I have a slight preference for moving this transformation logic into the magrittr pipeline above, assuming there isn't a good reason to mutate these fields in place. The goal would just be to keep things clean and consistent, so it's not necessary for correct functionality:

```R
agency_legacy_cw <-
  openxlsx::read.xlsx(...) %>%
  ...
  # Account for error in Clerk's report which lists Village of Skokie Library
  # Fund twice
  filter(!(agency_num == "031170001" & agency_num_24 == "031170000")) %>%
  # Correct error in Clerk's report which lists incorrect agency number for
  # the TIF VIL OF OLYMPIA FIELDS-GOV HWY/VOLL
  mutate(
    across(
      c(agency_num, agency_num_24),
      ~ if_else(
          agency_name == "TIF VIL OF OLYMPIA FIELDS-GOV HWY/VOLL",
          "030930502",
          .x
        )
    )
  )
```

data-raw/agency/agency.R
jeancochrane: That's right, sorry for the lack of clarity! I don't really know what we use these short names for, so I'll leave it up to you to decide if it's a problem that they're not included in the crosswalk.

jeancochrane (approved):
Alright, I think this is ready to go! I don't have 100% confidence that we got the data model exactly right, but I think it's good enough to move on to making code changes. Plus, we can always tweak this code later if we decide we want to change the data model.

data-raw/agency/agency.R
kyrasturgill: Agreed, I initially had code similar to what you had here but couldn't get it to work and was impatient - your code works though!

data-raw/agency/agency.R
kyrasturgill: I see... I am also not positive - I'm not totally confident as to how this crosswalk is going to work out so I imagine we'll likely need to revisit it. Hopefully we will have better clarity after testing it as to whether we need the other name iterations in the crosswalk.

PR #68, opened 2026-03-04
Update create_db

kyrasturgill:
Here are the proposed changes necessary for making the `create_db` script successfully run. Some of the tests in `create_db.sql` have been removed to account for oddities in the 2024 data. Ultimately I believe we should confirm with the Clerk's office / ask them to correct the source data rather than removing some of these tests but for the time being this is how I've gotten around to making a test database.

Other more necessary, permanent updates are:
* Add new field names for some of the agency and fund tables to `create_db.sql` which is used to construct the tables in the SQLite database
* Add new table `pin_tif_distribution` to `create_db.sql` and reference in `create_db.R`
* Some updates to the metadata for the database, but think we need to go through these / discuss updating the `DESCRIPTION` file


kyrasturgill:
Making this a draft PR as we know we still have edits to make such as:
- [x] update `DESCRIPTION` file and the database metadata that's derived from it
- [x] add back in tests that were removed to account for odd behavior in the source data - for discussion is whether the raw source data will be updated or if the ingest scripts should be updated 
- [ ] once the new database is ready for release, remove the pre-release tag

data-raw/create_db.R
```diff
@@ -73,10 +74,14 @@ desc_url_package <- desc %>%
   str_extract("(?<=URL: ).*(?=,)")
 
 db_base_url <- "https://ccao-data-public-us-east-1.s3.amazonaws.com/ptaxsim/"
-db_full_url <- paste0(db_base_url, "ptaxsim-", db_version, ".db.bz2")
+db_full_url <- paste0(
+  db_base_url, "ptaxsim-", db_version,
+  "-", db_version_label, ".db.bz2"
+)
```
kyrasturgill, line 88: Ok ultimately what I did here was causing the test failure because I made it so `db_version` would still not have any alpha characters but created this new object `db_version_label` so that the alpha tag would be included in the label and URL. I realize now this isn't ideal, but we'll need to change it when we create the final database anyway, but LMK if you'd prefer we tidy this up so the test databases technically have different versions.

kyrasturgill:
@jeancochrane - This is ready for your review!

data-raw/create_db.R
```diff
@@ -37,7 +37,8 @@ db_send_queries <- function(conn, sql) {
 # changes. This is checked against Config/Requires_DB_Version in the DESCRIPTION
 # file via check_db_version(). Schema is:
 # "MAX_YEAR_OF_DATA.MAJOR_VERSION.MINOR_VERSION"
-db_version <- "2023.0.0"
+db_version <- "2024.0.0"
+db_version_label <- "alpha.2" # informational only, not compared
```
jeancochrane: [Nitpick, optional] Thinking about future-proofing this solution: What do you think about renaming this variable to make it clearer that it's an optional pre-release identifier?

```R
# "MAX_YEAR_OF_DATA.MAJOR_VERSION.MINOR_VERSION-PRE_RELEASE_VERSION"
db_version <- "2024.0.0"
# Optional pre-release identifier. Informational only, not compared.
# Set this to an empty string for a public release, or to a string like 
# "alpha.1" for a release candidate
db_pre_release_version <- "alpha.2"
```

data-raw/create_db.R
jeancochrane: **[Suggestion, optional]** I think this is a nice workaround! My only thought is that it might be nice to preserve it for the future, since it's possible we'll want to publish pre-releases again in future years. I think we could do that pretty easily -- just make sure to swap in a new variable name for `db_version_label` if you take my suggestion above:

```R
db_full_url <- paste0(
  db_base_url, "ptaxsim-", db_version,
  if (nzchar(db_version_label)) paste0("-", db_version_label) else "",
  ".db.bz2"
)
```

data-raw/create_db.sql
```diff
@@ -4,35 +4,24 @@ PRAGMA foreign_keys = ON;
 CREATE TABLE agency (
     year                    int                                        NOT NULL,
     agency_num              varchar(9)                                 NOT NULL,
+    authority_num           varchar(6)                                         ,
```
jeancochrane: **[Suggestion, optional]** My understanding is that this field should always be non-null starting in 2024, right? What do you think about adding a `CHECK` constraint to that effect?

```sql
CHECK(year < 2024 OR authority_num IS NOT NULL)
```

If you like that idea, I think we could use a similar pattern to add `CHECK` constraints for the rest of the fields that are only present either pre- or post-2024.

data-raw/create_db.sql
```diff
@@ -4,35 +4,24 @@ PRAGMA foreign_keys = ON;
...
     home_rule_ind           boolean                                    NOT NULL,
     agg_ext_base_year       int      CHECK(agg_ext_base_year >= 2003)          ,
     lim_numerator           bigint   CHECK(lim_numerator >= 0)                 ,
     lim_denominator         bigint   CHECK(lim_denominator >= 0)               ,
     lim_rate                double   CHECK(lim_rate >= 0)                      ,
-    prior_eav               bigint   CHECK(prior_eav >= 0)             NOT NULL,
-    curr_new_prop           bigint   CHECK(curr_new_prop >= 0)         NOT NULL,
+    prior_eav               bigint   CHECK(prior_eav >= 0)                     ,
+    curr_new_prop           bigint                                     NOT NULL,
```
jeancochrane, line 16: **[Thought, non-blocking]** Clarifying this change for posterity: We learned that `curr_new_prop` can be negative if the value of disconnections (parcels that are no longer subject to the taxing authority of the agency) is greater than the value of new property.

data-raw/create_db.sql
```diff
@@ -66,16 +58,15 @@ CREATE TABLE agency_fund (
     agency_num              varchar(9)                                 NOT NULL,
     fund_num                varchar(3)                                 NOT NULL,
     levy                    bigint   CHECK(levy >= 0)                  NOT NULL,
-    loss_pct                double   CHECK(loss_pct >= 0 
-                                     AND   loss_pct <= 1)              NOT NULL,
```
jeancochrane, line 70: **[Question, non-blocking]** Remind me why `loss_pct` can be greater than 1?

data-raw/create_db.sql
```diff
@@ -66,16 +58,15 @@ CREATE TABLE agency_fund (
...
+    loss_pct                double   CHECK(loss_pct >= 0)              NOT NULL,
     levy_plus_loss          bigint   CHECK(levy_plus_loss >= 0)        NOT NULL,
-    rate_ceiling            double   CHECK(rate_ceiling >= 0)          NOT NULL, 
+    rate_ceiling            double   CHECK(rate_ceiling >= 0)          NOT NULL,
     max_levy                bigint   CHECK(max_levy >= 0)              NOT NULL,
     prelim_rate             double   CHECK(prelim_rate >= 0)           NOT NULL,
-    ptell_reduced_levy      bigint   CHECK(ptell_reduced_levy >= 0)            ,
-    ptell_reduced_ind       boolean                                    NOT NULL,
-    final_levy              bigint   CHECK(final_levy >= 0)            NOT NULL,
```
jeancochrane, line 77: **[Question, non-blocking]** Are we removing these checks because `ptell_reduced_levy` and `final_levy` can be less than 0? If so, do you recall why that is?

data-raw/create_db.sql
```diff
@@ -236,3 +231,31 @@ CREATE INDEX ix_tif_distribution_year_agency_num
...
+    pin_eav                 int      CHECK(pin_eav >= 0)               NOT NULL,
+    pin_frozen_eav          int      CHECK(pin_frozen_eav >= 0)        NOT NULL,
+    pin_revenue             double   CHECK(pin_revenue >= 0)           NOT NULL,
+    pin_increment_eav       int      CHECK(pin_increment_eav >= 0)     NOT NULL,
+    pin_distribution_pct    double   CHECK(pin_distribution_pct >= 0)  NOT NULL,
+    transit_tif_to_cps      double   CHECK(transit_tif_to_cps >= 0)            ,
+    transit_tif_to_tif      double   CHECK(transit_tif_to_tif >= 0)            ,
+    transit_tif_to_dist     double   CHECK(transit_tif_to_dist >= 0)           ,
+    is_transit_tif          boolean                                    NOT NULL,
+    PRIMARY KEY (year, pin, agency_num, tax_code_num)
```
jeancochrane: **[Question, optional]** Is `tax_code_num` necessary as part of this primary key? I would have guessed that the table would be unique by `(year, pin, agency_num)`. Not a big deal either way, I'm just curious to test my understanding.

data-raw/create_db.R
```diff
@@ -124,7 +130,9 @@ for (file in files) {
 datasets <- c("pin", "tax_code")
 for (dataset in datasets) {
   message("Now loading: ", dataset)
-  df <- collect(arrow::open_dataset(file.path(remote_bucket, dataset)))
+  df <- collect(arrow::open_dataset(file.path(remote_bucket, dataset),
+    unify_schemas = TRUE
```
jeancochrane, line 146: **[Suggestion, optional]** I remember this bug being kind of tricky, so it might be worth preserving what we learned in a comment:

```suggestion
    # Starting in 2024, there are some major changes regarding the columns
	# that are present in these data files. That means we need to unify the
	# schemas across files, since otherwise arrow will take the schema from
	# the first file it finds in the dataset
    unify_schemas = TRUE
```

jeancochrane (approved):
Nice work! A couple small suggested improvements below, but no major concerns.

data-raw/create_db.R
kyrasturgill: I like this! Added in.

data-raw/create_db.R
kyrasturgill: Added this, swapping `db_pre_release_version` for `db_version_label`.

data-raw/create_db.sql
kyrasturgill: Good thinking! This check has been added.

data-raw/create_db.sql
kyrasturgill: Ahh it should not be! I think it was because of an error with the agency ingest script where a problem fund was not being corrected because of our new `fund_num` format. So I rid this initially and forgot to add back in. I've added back in the check that it is not greater than 1.

data-raw/create_db.sql
kyrasturgill: There were a few bond funds that had -1 as values for these fields, but that has been corrected by the Clerk - I forgot to add these checks back in. Thank you for catching!

data-raw/create_db.sql
kyrasturgill: You're correct, it's not actually necessary. I think this must have been carried over from when I copied the code for `tif_distribution` table in which `tax_code_num` is a necessary key. I can remove `tax_code_num`.

data-raw/create_db.sql
```diff
@@ -66,14 +59,13 @@ CREATE TABLE agency_fund (
...
-    loss_pct                double   CHECK(loss_pct >= 0 
+    loss_pct                double   CHECK(loss_pct >= 0
                                      AND   loss_pct <= 1)              NOT NULL,
     levy_plus_loss          bigint   CHECK(levy_plus_loss >= 0)        NOT NULL,
-    rate_ceiling            double   CHECK(rate_ceiling >= 0)          NOT NULL, 
+    rate_ceiling            double   CHECK(rate_ceiling >= 0)          NOT NULL,
     max_levy                bigint   CHECK(max_levy >= 0)              NOT NULL,
     prelim_rate             double   CHECK(prelim_rate >= 0)           NOT NULL,
     ptell_reduced_levy      bigint   CHECK(ptell_reduced_levy >= 0)            ,
-    ptell_reduced_ind       boolean                                    NOT NULL,
```
kyrasturgill, line 76: This field is no longer present in the 2024 reports and was removed from the `agency_fund` table. I'd forgotten to remove it from the table schema here.

kyrasturgill:
@jeancochrane thanks for catching some of these checks I missed and forgot to add back in! I have addressed your comments. I also realized I had not removed `ptell_reduced_ind` which is no longer in the `agency_fund` table so I addressed that as well. If you can take another quick look then hopefully I am good to merge.

data-raw/create_db.R
```diff
@@ -33,11 +33,12 @@ db_send_queries <- function(conn, sql) {
 }
 
 
-# Set the database version. This gets incremented manually whenever the database
-# changes. This is checked against Config/Requires_DB_Version in the DESCRIPTION
-# file via check_db_version(). Schema is:
-# "MAX_YEAR_OF_DATA.MAJOR_VERSION.MINOR_VERSION"
-db_version <- "2023.0.0"
+# "MAX_YEAR_OF_DATA.MAJOR_VERSION.MINOR_VERSION-PRE_RELEASE_VERSION"
```
jeancochrane, line 39: **[Nitpick, required]** Oops, I think we may have lost some of the contextual prelude in the process of updating this comment:


```suggestion
# Set the database version. This gets incremented manually whenever the database
# changes. This is checked against Config/Requires_DB_Version in the DESCRIPTION
# file via check_db_version(). Schema is:
# "MAX_YEAR_OF_DATA.MAJOR_VERSION.MINOR_VERSION-PRE_RELEASE_VERSION"
```

data-raw/create_db.sql
```diff
@@ -4,35 +4,25 @@ PRAGMA foreign_keys = ON;
 CREATE TABLE agency (
     year                    int                                        NOT NULL,
     agency_num              varchar(9)                                 NOT NULL,
+    authority_num           varchar(6) CHECK(year < 2024
+                                       OR authority_num IS NOT NULL)           ,
```
jeancochrane, line 8: **[Suggestion, optional]** Thanks for adding the `year` check here! There are a couple more fields that I think could use a similar check to replace old `NOT NULL` checks:

- `prior_eav`
- `pct_burden`
- `agency_num_24`
- `agency_name_24`
- `exe_vet_dis_100`
- `tax_code_distribution_pct`

What do you think? Is it worth adding a similar type of year-based not-null check to these fields?

jeancochrane (approved):
Thanks for the quick turnaround! Just two small suggestions remaining.

data-raw/create_db.sql
kyrasturgill: Adding those checks makes sense to me - the only one that would not work is for `agency_num_24` and `agency_name_24`, as those do have NULL values in 2024 if the old agency did not have a change.

kyrasturgill: Oh and for `exe_vet_dis_100`, you had converted NULL values to 0 right? So we could just add a regular NOT NULL test back in for that one I think.

data-raw/create_db.sql
```diff
@@ -4,39 +4,33 @@ PRAGMA foreign_keys = ON;
...
     total_final_levy        bigint   CHECK(total_final_levy >= 0)      NOT NULL,
     total_final_rate        double   CHECK(total_final_rate >= 0)      NOT NULL,
-    reduction_type          varchar                                            ,
     reduction_pct           double   CHECK(reduction_pct >= 0
                                      AND   reduction_pct <= 1)                 ,
     total_non_cap_ext       double   CHECK(total_non_cap_ext >= 0)             ,
     total_ext               double   CHECK(total_ext >= 0)             NOT NULL,
+
+    CHECK(year >= 2024 OR prior_eav IS NOT NULL),
+    CHECK(year >= 2024 OR pct_burden IS NOT NULL),
```
kyrasturgill: I've added these checks here - so because in 2024, the Clerk report leaves these fields blank rather than in the past they would be 0, there will be NULL values in 2024. This test checks that any year less than 2024 will not have a NULL value. Did I do that right?

jeancochrane: **[Suggestion, optional]** If possible, I think it would be easier to read if we folded all the checks for these fields together, such that they're colocated with the field definitions. So for example, `prior_eav` is currently defined like this:

```sql
prior_eav               bigint   CHECK(prior_eav >= 0)                     ,
```

And then the table has this check defined at the bottom:

```sql
    CHECK(year >= 2024 OR prior_eav IS NOT NULL),
```

But I think in theory we could just combine the two:

```sql
    prior_eav               bigint   CHECK(prior_eav >= 0)
                                     AND (year >= 2024 OR prior_eav IS NOT NULL),
```

Does that make sense? Not a huge deal either way I don't think.

jeancochrane (commented):
One small thought about colocating the checks with the field definitions! Let me know if it would be easier to chat this out. I'm not committed to the idea either, just a suggestion for maximum clarity.

data-raw/create_db.sql
kyrasturgill: I tried doing that initially but I get an error! Due to the syntax I believe. I consulted Copilot which recommended the method I went with where it's separated out as I have it now. That ok?

jeancochrane: Do you mind sharing the error you got? Probably fine to do it this way, but I'm curious if it's a syntax error that we could overcome.

kyrasturgill: Error in `map()`:
ℹ In index: 2.
Caused by error:
! near "AND": syntax error

kyrasturgill: Actually this may work:
`    prior_eav               bigint   CHECK((prior_eav >= 0)
                                     AND (year >= 2024 OR prior_eav IS NOT NULL)),`

jeancochrane (approved):
Nice work!

PR #69, opened 2026-03-19
Code base changes to PTAXSIM for 2024 update

kyrasturgill:
This PR makes following necessary changes for PTAXSIM code in order for the package to function with the updates to the 2024 data structure:
- Add `exe_vet_dis_100` to `lookup_pin()` 
- Create a new lookup function, `lookup_pin_tif()` which queries the table that now contains TIF distribution share per PIN
- Add a pre-2024 and 2024 onward logic for joining TIF distribution share to the `pin_dt` (pre-2024 joins `tif_dt` by tax code, whereas post 2024 joins by `pin_tif_dt`)


R/lookup.R
```diff
@@ -393,3 +394,51 @@ lookup_tif <- function(year, tax_code, conn = ptaxsim_db_conn) {
 
   return(dt)
 }
+
+#' @describeIn lookup_dt Lookup any PINs within a TIF
+#'   year. Returns a \code{data.table} with only 1 row per PIN and year,
+#'   or 0 rows if the PIN is not in a TIF.
```
jeancochrane: **[Question, non-blocking]** Am I right that this lookup will only return values post-2024, since pre-2023 PINs do not have entries in the `pin_tif_distribution` table? If so, should we leave a note about that in this docstring, or would you rather leave it up to the TIF vignette to explain that nuance?

jeancochrane (approved):
Looks great! For the sake of test coverage, at some point we should add some tests for `lookup_pin_tif()` and the post-2024 TIF join in `tax_bill()`. However, I'm fine punting that until later, since this interface is still in the experimental phase and we want to get it out for testing ASAP. If we do decide to punt adding new tests, let's just make sure to add a checklist item to the "unit/integration tests" section of https://github.com/ccao-data/ptaxsim/issues/59 to remind ourselves to add tests for any new code paths before the final 2024 release.

R/lookup.R
kyrasturgill: That's a good point - I think adding a note here and to `lookup_tif` about the 2024 split makes sense. I can do that now.

PR #70, opened 2026-03-19
Update configs and workflows to point to 2024 pre-release DB for testing and documentation

jeancochrane:
This PR makes a few changes on top of https://github.com/ccao-data/ptaxsim/pull/69 to get a few tricky tests passing:

- Updates `Config/Wants_DB_Version` in the `DESCRIPTION` file to point to the pre-release version of the 2024 database, so that our GitHub workflows will pull this version of the database for testing
- Reknits the README so that it includes references to the 2024.0.0 database, which is necessary in order to get tests passing that check to ensure the README DB version matches the `DESCRIPTION` file
- Tweaks the test that checks the output columns for the `lookup_pin()` function to make sure it expects the new `exe_vet_dis_100` column


.github/actions/prepare-ptaxsim/action.yaml
```diff
@@ -31,7 +31,7 @@ runs:
       id: version_db
       run: |
         DESCRIPTION_PATH=$(echo "${{ github.workspace }}/DESCRIPTION" | sed 's/\\/\//g')
-        echo "PTAXSIM_VERSION=$(sed -n 's/.*Wants_DB_Version: \([0-9]*\.[0-9]*\.[0-9]\).*/\1/p' $DESCRIPTION_PATH)" >> $GITHUB_ENV
+        echo "PTAXSIM_VERSION=$(sed -n 's/.*Wants_DB_Version: \([0-9]*\.[0-9]*\.[0-9][0-9A-Za-z.-]*\).*/\1/p' $DESCRIPTION_PATH)" >> $GITHUB_ENV
```
jeancochrane, line 34: We need to update this regex to account for the possibility of a pre-release suffix like `-alpha.1`.

man/figures/README-multi_year_4-1.png
jeancochrane, line 1: I'm actually not entirely sure why reknitting the README generated a diff in this file. Do you know? It seems like the final TIF amount is indeed slightly different. In any case, I'm not too worried about it since I assume we'll need to tweak this code later on as part of the 2024 release anyway.

README.md
```diff
@@ -640,7 +640,7 @@ multiple_years_plot <- ggplot(data = multiple_years_summ) +
 
 </details>
 
-<img src="man/figures/README-multi_year_4-1.png" width="85%" />
+<img src="man/figures/README-multi_year_4-1.png" alt="" width="85%" />
```
jeancochrane, line 643: I'm also not sure why reknitting the README made this change, but it doesn't seem like a big deal.

man/figures/README-multi_year_4-1.png
kyrasturgill: No idea.. I thought this image was dropped in manually too? But yes I'll look into whether/how we update this visual in the README.

kyrasturgill (approved):
Thanks for this! All looks good to me.

PR #79, opened 2026-03-26
Update sample tax bills for 2024

jeancochrane:
This PR updates our set of sample tax bills in the `data-raw/sample_tax_bills/` subdirectory to include sample bills for 2024. It builds off of https://github.com/ccao-data/ptaxsim/pull/78, since sample tax bills are an important part of some unit tests.


data-raw/sample_tax_bills/agency_name_match.csv
```diff
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:6cb20066e916e3fe3650adc440717efd02c618c317102622c3d74e59ba2d0eac
-size 9768
+oid sha256:659f8e8f17bc5ef5fd6474e984b62deb480da8e9d1ebad4d73733726b65a83aa
```
jeancochrane, line 2: This file needed to change to add some agency names that are new or modified as of 2024. The list of new agencies is likely longer than it was in the past due to the way that the Clerk changed the format for so many agency names during the ias transition. In the code below, I leave a detailed comment explaining how to update this file for future years.

data-raw/sample_tax_bills/sample_tax_bills_detail.csv
```diff
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:7b6c9404189af79c884ffb3af225c1122bc911189235b9f16b5119f6c67a0029
-size 69493
+oid sha256:564ec164b1dc073500912afa16e6e524eedff5f2679e275893c78ce945c53c07
```
jeancochrane, line 2: This file got changed automatically upon running the `data-raw/sample_tax_bills/sample_tax_bills_detail.R` script with the changes below.

data-raw/sample_tax_bills/sample_tax_bills_detail.R
```diff
@@ -64,7 +67,8 @@ extract_tax_bill <- function(file) {
         paste0(
           "TAXES|Assess|Property|EAV|Local Tax|",
           "Total Tax|Do not|Equalizer|cookcountyclerk.com|",
-          "Pursuant|meaning of|If paying later|\\d{15}+|By \\d{2}/"
+          "Pursuant|meaning of|If paying later|\\d{15}+|By \\d{2}/|",
+          "Visit COOKCOUNTYCLERKIL.GOV"
```
jeancochrane, line 71: This line appears to be present on bills as of 2024, and we need to remove it:

Image

data-raw/sample_tax_bills/sample_tax_bills_detail.R
```diff
@@ -94,28 +97,63 @@ bills_df <- bills_df %>%
...
+#          the code below will ensure we roll up those fund values into the
+#          overall values for the agency.
+#
+#   2. agency_num: The Clerk's unique identifier for the agency. You can find
+#      this value in the Clerk's agency rate report and/or TIF report.
+#
+#   3. name_priority: Integer priority for this name, in descending priority
+#      order (e.g. 1 is higher priority than 2). We only use this value when
+#      rolling up funds and transit TIF distributions into their parent agency.
+#      In these cases, it's important that the parent agency have priority 1.
```
jeancochrane: Am I using "parent agency" correctly here, and elsewhere in the PR?

data-raw/sample_tax_bills/sample_tax_bills_detail.R
```diff
@@ -94,28 +97,63 @@ bills_df <- bills_df %>%
...
+# Consolidate funds and TIF breakouts into a single line-item for the parent
+# agency
 bills_df <- bills_df %>%
-  mutate(cook = str_detect(
-    agency_name,
-    "Cook County Public Safety|Cook County Health Facilities|County of Cook"
-  )) %>%
-  group_by(pin, year, cook) %>%
-  mutate(across(final_tax:prev_tax, ~ ifelse(cook, sum(.x), .x))) %>%
-  ungroup() %>%
```
jeancochrane, line 115: This is no longer necessary because I've added these Cook County agencies to the `agency_name_match` crosswalk, meaning we can treat them the same way we treat TIFs and other agencies with `name_priority` values other than 1.

tests/testthat/_snaps/lookup_agency_over_time/lookup.md
```diff
@@ -1,4 +1,4 @@
-# lookup values/data are correct
+# agency lookup matches snapshots
```
jeancochrane, line 1: I split this snapshot test out into its own `testthat` test, with a dedicated test name. That test name got automatically updated here as a result. See [these docs](https://testthat.r-lib.org/reference/expect_snapshot_value.html) for a refresher on snapshot tests.

tests/testthat/_snaps/lookup_agency_over_time/lookup.md
```diff
@@ -68,3 +68,4 @@
         }
       ]
     }
+
```
jeancochrane, line 71: `testthat` likes to add trailing newlines when creating snapshots. In https://github.com/ccao-data/ptaxsim/pull/64, pre-commit stripped these out automatically due to [the `end-of-file-fixer` hook](https://github.com/ccao-data/ptaxsim/blob/eb1e3e8d4c7426656ea38b78d9d502b2f06286d2/.pre-commit-config.yaml#L25-L26). In the diff below, you'll find that I've tweaked this hook definition in `.pre-commit-config.yml` to exclude snapshot files like this one, so `testthat` can use the trailing newline format it prefers for these files. This follows the same pattern that we introduced in the `ccao` package in https://github.com/ccao-data/ccao/pull/51.

tests/testthat/test-lookup.R
```diff
@@ -286,7 +286,9 @@ test_that("lookup values/data are correct", {
       3178945619, 286280738, 0, 638172798
     )
   )
+})
 
+test_that("agency lookup matches snapshots", {
```
jeancochrane, line 294: It was annoying to debug these snapshot tests because they were getting bundled in with the simpler `expect_equal` tests immediately above, so I split them out into their own dedicated test.

tests/testthat/test-lookup.R
```diff
@@ -298,6 +300,7 @@ test_that("lookup values/data are correct", {
     variant = "lookup_agency_over_time"
   )
   expect_snapshot_value(
+    # We expect this to change every year as we add new bills to the summary DF
```
jeancochrane: This isn't the best design, since tests should ideally only fail if something is wrong with new data. I'd rather not deal with refactoring it right now, however.

tests/testthat/test-sample_tax_bills_summary.R
```diff
@@ -0,0 +1,36 @@
+context("Test sample_tax_bills_summary")
```
jeancochrane, line 1: Now that I understand that the sample tax bills process is manual, I added some simple tests here to make sure we update them properly in future years.

.pre-commit-config.yaml
```diff
@@ -23,7 +23,7 @@ repos:
     -   id: file-contents-sorter
         files: '^\.Rbuildignore$'
     -   id: end-of-file-fixer
-        exclude: '\.Rd|\.csv'
+        exclude: '(\.Rd|\.csv|^tests/testthat/_snaps/.*\.md$)'
```
jeancochrane, line 26: As mentioned above, this change is necessary in order to prevent pre-commit from stripping trailing newlines from testthat snapshot files.

tests/testthat/test-tax_bill.R
```diff
@@ -131,84 +132,196 @@ test_that("grid expansion works correctly", {
...
-exclude_pins <- c("20031180060000")
-sum_dt <- sum_dt %>%
+exclude_pins <- c(
+  # Bill lists different pre- and post-exemption amounts, but does not show
+  # any exemptions, and the Clerk data also has no exemptions
+  "20031180060000",
+  # Has a 2022 vetdis 100% exemption that doesn't seem to be present in the
+  # Clerk data
+  "10252080490000"
+)
```
jeancochrane, line 142: We were previously excluding this PIN from tests lower down in the file. I'm moving it up here to better centralize the list of PINs that we want to exclude from testing due to weird edge cases.

tests/testthat/test-tax_bill.R
```diff
@@ -131,84 +132,196 @@ test_that("grid expansion works correctly", {
...
+# Get a list of PINs that are in transit TIFs. We exclude these from our
+# bill summary tests, because the Treasurer's method for calculating transit
+# TIF distributions does not match our method
+transit_tif_pins <- sum_tax_bill_simplified %>%
+  select(year, pin, agency_num, final_tax) %>%
+  count(year, pin, agency_num) %>%
+  # PINs in transit TIFs will have two line items with the CPS agency number,
+  # one being the TIF distribution
+  filter(n > 1) %>%
+  pull(pin)
```
jeancochrane, line 168: This operation is also not new, but is rather moved up from a spot further down. I've also renamed the variable from `transit_bills` to `transit_tif_pins` for clarity.

tests/testthat/test-tax_bill.R
```diff
@@ -131,84 +132,196 @@ test_that("grid expansion works correctly", {
...
+    group_by(year, pin, agency_num) %>%
+    summarize(
+      agency_name = first(agency_name),
+      final_tax = sum(final_tax, na.rm = TRUE),
+      .groups = "drop"
+    ) %>%
+    # Convert output to dataframe because it is the simplest possible data
+    # structure for the purposes of comparison
+    as.data.frame()
+}
```
jeancochrane, line 199: I'm curious to get your feedback on this approach! I'm not 100% confident that I'm doing this correctly, or that we should even do it at all. Here's a motivating example explaining why rolling up agencies is important to get the bill comparison tests passing:

#### Bill calculated using `tax_bill` (Clerk data)

```R
all_bills_actual %>%
  filter(pin == '11302010040000', str_detect(agency_num, "^03038")) %>%
  select(year, pin, agency_num, agency_name, agency_tax_rate, final_tax)
```

```
    year            pin agency_num      agency_name agency_tax_rate final_tax
   <int>         <char>     <char>           <char>           <num>     <num>
1:  2024 11302010040000  030380000 CITY OF EVANSTON      0.01519095   1047.01
```

#### Bill extracted from Treasurer PDF

```R
all_bills_expected %>%
  filter(pin == '11302010040000', str_detect(agency_num, "^03038")) %>%
  select(year, pin, agency_num, agency_name, rate, final_tax)
```

```
   year pin            agency_num agency_name                          rate final_tax
  <int> <chr>          <chr>      <chr>                               <dbl>     <dbl>
1  2024 11302010040000 030380000  CITY OF EVANSTON                    1.27      874. 
2  2024 11302010040000 030380001  CITY OF EVANSTON LIBRARY FUND       0.231     159. 
3  2024 11302010040000 030380002  CITY OF EVANSTON GENERAL ASSISTANCE 0.02      13.8
```

Note that the rate and tax totals match up when you roll up these Treasurer agencies. Hence, this function groups agencies by the first 8 digits of the agency num (everything before the last digit) and selects the agency with the lowest `agency_num` as the "parent" (in practice, this winds up being the agency whose number ends in 0). That seemed reasonable to me, but I'm curious what you think as someone who has had more experience with the new fund reporting -- is there a more comprehensive way to resolve these discrepancies using e.g. the agency crosswalk?

tests/testthat/test-tax_bill.R
```diff
@@ -131,84 +132,196 @@ test_that("grid expansion works correctly", {
...
+
+test_that("returns correct row counts for all sample bills", {
   expect_equal(
-    tax_bill(sum_dt$year, sum_dt$pin, simplify = FALSE) %>%
-      nrow(),
-    753
```
jeancochrane, line 147: I found these raw counts to be unhelpful, since we need to update them every time we add new sample bills, and there's no theoretical basis indicating whether they are correct. My change switches to comparisons that I find clearer and that shouldn't have to change every time we add new sample bills.

tests/testthat/test-tax_bill.R
```diff
@@ -131,84 +132,196 @@ test_that("grid expansion works correctly", {
...
-test_that("all differences are less than $25", {
-  expect_true(
-    left_join(
-      tax_bill(sum_dt_no_rpm$year, sum_dt_no_rpm$pin) %>%
-        select(year, pin, agency_num, tax_calc = final_tax),
-      det_dt %>%
-        select(year, pin, agency_num, tax_real = final_tax) %>%
-        as_tibble(),
-      by = c("year", "pin", "agency_num")
+test_that("amounts correct for non-transit-TIF sample bills", {
```
jeancochrane, line 222: The diff is hard to parse, but this is the new version of the test that was previously named `returned amount/output correct for all sample bills`.

tests/testthat/test-tax_bill.R
```diff
@@ -131,84 +132,196 @@ test_that("grid expansion works correctly", {
...
+    ) %>%
+    select(-agency_name_expected, -agency_name_actual) %>%
+    relocate(agency_name, .after = agency_num) %>%
+    filter(if_any(starts_with("chk_"), ~ .x == TRUE))
+
+  expect_equal(nrow(bill_mismatches), 0)
 })
 
+test_that(
+  "all diffs are less than $25 for non-transit-TIF sample bills pre-2024",
```
jeancochrane, line 266: This diff is also hard to read, but this is the new version of the test that was previously named `all differences are less than $25`. I also split that test out into a pre-2024 version (this one) and a post-2024 version (next one) so that we can handle the agency fund reporting issue post-2024.

tests/testthat/test-tax_bill.R
```diff
@@ -272,7 +385,7 @@ test_that("Returns 0 for agency with base/levy of 0", {
 
 test_that("Simplify FALSE / TRUE identical", {
   # transit tif tax code
-  transit_tif_pins <- DBI::dbGetQuery(
+  rpm_tif_pins <- DBI::dbGetQuery(
```
jeancochrane, line 392: Now that we're using the variable name `transit_tif_pins` above to refer to the PINs that should be excluded from comparison tests, this variable needs a different name. I think `rpm_tif_pins` is more appropriate anyway, since this query only filters for the RPM TIF, excluding RLE.

data-raw/sample_tax_bills/sample_tax_bills_detail.R
```diff
@@ -94,28 +97,63 @@ bills_df <- bills_df %>%
   filter(!stringr::str_detect(agency_name, " Total")) %>%
   mutate(across(c(year, final_tax:prev_tax), readr::parse_number))
 
-# Load agency name lookup from file
+# Load agency name lookup from file. This lookup maps agency names to numbers.
+#
+# We maintain this file by hand. When adding a new year of sample bills, you
+# will likely encounter agencies that are not yet present in this list. To
```
kyrasturgill: **[nitpick, non-blocking]** I wonder if this is likely in future years, when it's not as weird as 2024? Maybe this could be "it is possible"?

data-raw/sample_tax_bills/sample_tax_bills_detail.R
kyrasturgill: This makes sense to me! The only thing I was slightly confused by was the transit TIF having a "parent" agency. For those cases, is the fund with priority 1 just the name of the transit TIF (no specification of CPS or other agency distribution?)

tests/testthat/test-lookup.R
kyrasturgill: This specific test I think I just need to talk through with you to make sure I understand, but I trust your judgement here!

tests/testthat/test-sample_tax_bills_summary.R
kyrasturgill: Good thinking!

tests/testthat/test-tax_bill.R
kyrasturgill: I like this approach a lot actually! It is more intuitively clear to me than the agency crosswalk. The only flag I'd raise is if this pattern does not for some reason hold true in the future - perhaps another question we could pose to the Clerk?
But as long as it's working for the current agency structure, then I think this approach is worth keeping!

tests/testthat/test-tax_bill.R
kyrasturgill: Makes sense to me!

tests/testthat/test-tax_bill.R
```diff
@@ -131,84 +132,200 @@ test_that("grid expansion works correctly", {
...
-      nrow(),
-    753
+    # We're not sure if this will continue to hold after 2024, when the
+    # Treasurer started reporting a slightly different set of agencies on
+    # tax bills compared to the Clerk's data (largely driven by the Clerk's
+    # decision to split some funds out into their own agencies, though there
+    # are other inconsistencies). It may just be a coincidence that the counts
+    # match in 2024. If this is the case, we should switch to an approach that
+    # rolls up funds into their parent agency, and then compares the row counts
+    # for parent agencies only
```
kyrasturgill, line 210: I think this sums up perfectly the flag raised in my comment above - I think this documentation is great for sake of making sure to confirm this approach holds in the future.

kyrasturgill (approved):
This is incredible work! I don't really have much to add here, besides affirming that I think your parent agency roll-up method makes sense. I echo your concern about that pattern maybe not holding in future years but if it works this year, then I think it's fine to keep it. 
Thank you for slogging through these changes!

tests/testthat/test-tax_bill.R
jeancochrane: OK cool! Out of curiosity, do you think we could use the crosswalk to accomplish this same goal? If so, I think it might be preferable, just so that we can stay consistent with the abstractions  that we use to handle this type of problem.

data-raw/sample_tax_bills/sample_tax_bills_detail.R
jeancochrane: Good point! Tweaked in 6df55bb98d0cb8f12e4e12c815ed305321c6ecdc.

data-raw/sample_tax_bills/sample_tax_bills_detail.R
jeancochrane: That's a great question! I believe your summary is correct. To verify, here are all the rows containing transit TIFs (I added the agencies with all caps names, which is the new 2024 format):

| Agency Name | Agency Number | Priority |
| --- | --- | --- |
| City of Chicago - TIF Transit RPM1 | 030210900 | 1 |
| TIF TRANSIT CITY OF CHICAGO-RPM1 | 030210900 | 1 |
| City of Chicago - TIF RPM1 Distribution | 030210900 | 2 |
| Board of Education - from Transit TIF | 030210900 | 2 |
| Board of Education - TIF Transit RPM1 | 030210900 | 3 |
| BOARD OF EDUCATION - TIF RPM1 | 030210900 | 3 |

I tweaked the docs to clarify this point in ad73e8ba971bda5eb4548f1d432c3b466f12dc5d.

tests/testthat/test-lookup.R
jeancochrane: We talked this through in person and decided that it would be better to restrict the set of years for this test so that it doesn't fail every time we add new years of sample bills. I updated the test to do that in 7b3104dcee3c461e68f345a3f43ab4141af470bd and added an extra comment clarifying the purpose of these snapshot tests.

jeancochrane (commented):
Thanks for the review @kyrasturgill! I resolved all comments that don't need further discussion. I think the only remaining question is whether it would be better to use the agency crosswalk in place of the `rollup_agencies()` function in the test, so as to use a consistent interface for comparing agencies across years. I _think_ that is a desirable change for us to make, but I'm curious to get your take as a test of my understanding.

jeancochrane:
Replying to my question from above:

> I think the only remaining question is whether it would be better to use the agency crosswalk in place of the `rollup_agencies()` function in the test, so as to use a consistent interface for comparing agencies across years.

After poking around at it, I think this is actually not a feasibile idea. Though some of the mismatches could be resolved with the crosswalk, ~49 of the mismatches are for years before 2024, so the agency-to-fund change is not the primary driver of these differences.

tests/testthat/test-tax_bill.R
jeancochrane: I actually don't think the crosswalk is the right fit for this operation; see https://github.com/ccao-data/ptaxsim/pull/79#issuecomment-4337331024.

PR #80, opened 2026-04-02
Additional updates to agency ingest script

kyrasturgill:
This PR makes a few tweaks to the `agency.R` ingest script. After some consideration, we've opted to remove the individual collar county EAV fields (i.e. `dupage_cty_eav`) given these are no longer included as individual fields in the 2024 agency reports. Rather, they are summed to `overlap_eav`. This will resolve the issue that persists with the 2021 agency report in which two county EAV fields are mislabeled. 

The other change is revising the table `agency_fund_info`. This table will now include `agency_num`, given the discover that different agencies with the same `fund_num` have different fund names (mainly Bond Series funds). I also created a field `fund_type_name`, which pulls the standard name for a fund prior to 2024. This will allow a user to more easily analyze funds aggregated by their broader categories. 

@jeancochrane  I know this is a bit more involved than what we discussed, so let me know if you want to talk through or have any concerns with this approach!!

### DIFF
--- a/data-raw/agency/agency.R
+++ b/data-raw/agency/agency.R
@@ -294,19 +294,8 @@ agency <- map_dfr(file_names, function(file) {
     home_rule_ind = home_rule_ind %in% c("Y", "HR", "No PTELL"),
     home_rule_ind = replace_na(home_rule_ind, FALSE),
     cty_overlap_eav = ifelse(year < "2024",
-      rowSums(across(all_of(c(
-        "cty_dupage_eav",
-        "cty_lake_eav",
-        "cty_will_eav",
-        "cty_kane_eav",
-        "cty_mchenry_eav",
-        "cty_dekalb_eav",
-        "cty_kankakee_eav",
-        "cty_kendall_eav",
-        "cty_grundy_eav",
-        "cty_lasalle_eav",
-        "cty_livingston_eav"
-      )))),
+      rowSums(across(collar_counties
+      )),
       cty_overlap_eav
     ),
     across(
@@ -600,19 +589,8 @@ agency_legacy_cw <-
   unique() %>%
   # Account for error in Clerk's report which lists Village of Skokie Library
   # Fund twice
-  filter(!(agency_num == "031170001" & agency_num_24 == "031170000")) %>%
-  # Correct error in Clerk's report which lists incorrect agency number for
-  # the TIF VIL OF OLYMPIA FIELDS-GOV HWY/VOLL
-  mutate(
-    across(
-      c(agency_num, agency_num_24),
-      ~ if_else(
-        agency_name_24 == "TIF VIL OF OLYMPIA FIELDS-GOV HWY/VOLL",
-        "030930502",
-        .x
-      )
-    )
-  )
+  filter(!(agency_num == "031170001" & agency_num_24 == "031170000"))
+
 
 agency_info <- agency_info %>%
   left_join(agency_legacy_cw, by = "agency_num") %>%
--- a/data-raw/agency/agency.R
+++ b/data-raw/agency/agency.R
@@ -197,18 +197,18 @@ arrow::write_parquet(
 # Define collar county EAV fields
 collar_counties <-
   c(
-  "cty_dupage_eav",
-  "cty_lake_eav",
-  "cty_will_eav",
-  "cty_kane_eav",
-  "cty_mchenry_eav",
-  "cty_dekalb_eav",
-  "cty_kankakee_eav",
-  "cty_kendall_eav",
-  "cty_grundy_eav",
-  "cty_lasalle_eav",
-  "cty_livingston_eav"
-)
+    "cty_dupage_eav",
+    "cty_lake_eav",
+    "cty_will_eav",
+    "cty_kane_eav",
+    "cty_mchenry_eav",
+    "cty_dekalb_eav",
+    "cty_kankakee_eav",
+    "cty_kendall_eav",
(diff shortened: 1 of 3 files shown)


data-raw/agency/agency.R
jeancochrane: **[Nitpick, optional]** Could we reuse `collar_counties` here, for the sake of centralization? This code is untested and speculative, so feel free to disregard if it's not practical:

```suggestion
      rowSums(across(all_of(collar_counties))),
```

data-raw/agency/agency.R
```diff
@@ -152,11 +152,19 @@ agency_fund <- map_dfr(file_names, function(file) {
...
+  group_by(fund_type_num) %>%
+  summarise(fund_type_name = calc_mode(fund_name))
+
+
 # Breakout the fund names into their own table
 agency_fund_info <- agency_fund %>%
-  group_by(fund_type_num, fund_num) %>%
+  group_by(agency_num, fund_type_num, fund_num) %>%
   summarise(fund_name = calc_mode(fund_name)) %>%
   ungroup() %>%
```
jeancochrane, line 166: **[Question, non-blocking]** Since `agency_fund` is already unique by `agency_num` and `fund_num` (IIRC), do we still need to summarize the `fund_name` and set it to its mode? Or is my intuition correct that there should only ever be one `fund_name` for each `agency_num`/`fund_num` pair in the base `agency_fund` table to start with?

jeancochrane (approved):
Nice and simple, thanks! Just a couple potential simplifications below.

data-raw/agency/agency.R
kyrasturgill: I did have this using `collar_counties` originally but tidyselect returns a warning - per "[using an external vector in selections is ambiguous](https://tidyselect.r-lib.org/reference/faq-external-vector.html)" so I just reverted to naming them all. We could omit the `all_of()` helper function, since we're pretty confident there won't be any new collar counties added to the data any time soon?

data-raw/agency/agency.R
kyrasturgill: Theoretically that should be true - but I believe my instinct to do this was from there being cases where the same funds have a `fund_name` with slightly different spelling or typos across years. I think that was the intention with creating the `calc_mode` function.

data-raw/agency/agency.R
```diff
@@ -224,6 +248,9 @@ agency <- map_dfr(file_names, function(file) {
     rename_with(~"reduction_pct", any_of(c(
       "reduction_percent", "reduction_factor", "clerk_reduction_factor"
     ))) %>%
+    rename_with(~"total_non_cap_ext", any_of(c(
+      "total_non_cap_ext", "final_non_cap_ext", "total_non_cap_extension"
+    ))) %>%
```
kyrasturgill, line 253: I am not sure why but I mistakenly documented that `total_non_cap_ext` was no longer present in the 2024 report. This is not the case! This actually seems like it could be an important field for PTELL analysis, and there's no reason to no longer include it, so I'm adding it back in here.

data-raw/agency/agency.R
jeancochrane: That makes sense, thanks!

data-raw/agency/agency.R
jeancochrane: Nice catch!

data-raw/agency/agency.R
```diff
@@ -315,12 +331,12 @@ agency <- map_dfr(file_names, function(file) {
...
       c(
         lim_rate, pct_burden, total_prelim_rate, total_final_rate,
-        reduction_pct, total_ext
+        reduction_pct, total_non_cap_ext, total_ext
       ),
       ~ as.double(.x)
     )
   ) %>%
-  select(-total_reduced_levy) %>%
+  select(-total_reduced_levy, -collar_counties) %>%
```
jeancochrane: **[Nitpick, required]** I think there may be a bug here: As it's written now, I believe the dplyr syntax `select(-collar_counties)` is going to try to remove a column named `collar_counties` from the dataframe, rather than perform the operation that I think we intend, which is to expand out the `collar_counties` vector and remove each of its elements from the dataframe columns. For correct behavior, I think we need to wrap `collar_counties` in either `all_of()` (if we know every one of these columns will always be present in the dataframe at this point) or `any_of()` (if it's possible that some columns may not be present at this point).

jeancochrane (approved):
I like the further simplifications! Once instance of what I think is a bug, but otherwise, this looks great.

data-raw/agency/agency.R
jeancochrane: Hmm, I'm not sure I understand what you mean! I suspect the warning about using external vectors in selection is related to the `select()` call that I commented on below -- if we don't use `all_of()` or `any_of()`, tidyselect might interpret `collar_counties` as a column name rather than a vector that needs expansion. We definitely need either `all_of()` or `any_of()` here -- `all_of()` is appropriate if we're confident these columns will all always be present in the dataframe, otherwise `any_of()` is the right choice. Am I following you correctly? Happy to talk this out if it would be faster to clarify things.

kyrasturgill: Ohhh I completely confused myself -- I think the warning was from the later use of `select(-collar_counties)` which you noted. Not sure what was going on in my head.. but anyway this I'll keep the same and I've fixed the later issue where I did not use `select(all_of())`

data-raw/agency/agency.R
kyrasturgill: Fixed!

PR #84, opened 2026-04-16
2024 agency crosswalk vignette

### LINKED ISSUE #73: Add vignette demonstrating how to use the 2024 agency crosswalk
A number of agency numbers changed in 2024 due to changes in the way the Clerk tracks funds in their data. To track these agencies over time, we're adding a crosswalk that maps 2024 agency numbers to their pre-2024 iteration. It would be useful to document how to use this crosswalk, but we don't think it fits neatly into any of the existing vignettes, so let's make a new vignette showing how it works.

kyrasturgill:
I want to get this up sooner than later so you can start a review if you'd like! I still need to add a bit of narrative at the end of to tie it together. I was debating adding in a section about EAV and new property, but didn't want to stray too far from the original goal of just showing how to use the agency crosswalk. Open to general initial feedback on the structure and content first!

### DIFF
--- a/vignettes/agencies.Rmd
+++ b/vignettes/agencies.Rmd
@@ -15,15 +15,15 @@ knitr::opts_chunk$set(
 
 One of the unique features of PTAXSIM is the database that accompanies the package - this DB contains the only publicly accessible, machine-readable data collected from the various Cook County property tax agencies going back to 2006. While the primary use of PTAXSIM is its functionality to calculate tax bills, its database holds a plethora of data that can be used to investigate Cook County's property tax system, including the behavior of the over 900 taxing agencies and 400 TIFs that collect revenue through property taxes.
 
-In this vignette, we'll demonstrate how to query data from the PTAXSIM SQLite database that can help us analyze taxing agencies' property tax revenue over time, as well as revenue collected by TIFs.
+In this vignette, we'll demonstrate how to query data from the PTAXSIM SQLite database that can help us analyze taxing agencies' property tax revenue over time.
 
-Additionally, we'll show how to account for the 2024 changes to the Cook County Clerk's agency fund report data structure which will be necessary for users to pay attention to when conducting a time series analysis of certain taxing agencies.
+Additionally, we'll show how to account for the 2024 changes to the Cook County Clerk's agency data structure which will be necessary for users to account for when conducting a time series analysis of certain taxing agencies.
 
 # Chicago taxing agencies
 
-Using data from the PTAXSIM database, let's look at the levy history for the primary taxing agencies that appear on a Chicago property's tax bill: the `CITY OF CHICAGO`, `BOARD OF EDUCATION` (a.k.a. Chicago Public Schools), and `CHICAGO PARK DISTRICT`.
+Using data from the PTAXSIM database, let's look at the levy history for three of the primary taxing agencies that appear on a Chicago property's tax bill: the `CITY OF CHICAGO`, `BOARD OF EDUCATION` (a.k.a. Chicago Public Schools), and `CHICAGO PARK DISTRICT`.
 
-First, load some useful libraries and instantiate a PTAXSIM DBI connection with the default name (`ptaxsim_db_conn`) expected by PTAXSIM functions.
+First, load some useful libraries and instantiate a PTAXSIM DBI connection with the default name (`ptaxsim_db_conn`).
 
 ```{r}
 library(data.table)
@@ -34,18 +34,45 @@ library(ggplot2)
 library(ptaxsim)
 library(tidyr)
 
-#ptaxsim_db_conn <- DBI::dbConnect(RSQLite::SQLite(), here("./ptaxsim.2024.0.0-alpha.2.db"))
+ptaxsim_db_conn <- DBI::dbConnect(RSQLite::SQLite(), here("./ptaxsim.2024.0.0-alpha.2.db"))
 ```
 
 ```{r, echo=FALSE}
 # This is needed to build the vignette using GitHub Actions
-ptaxsim_db_conn <- DBI::dbConnect(
-  RSQLite::SQLite(),
-  Sys.getenv("PTAXSIM_DB_PATH")
-)
+#ptaxsim_db_conn <- DBI::dbConnect(
+#  RSQLite::SQLite(),
+#  Sys.getenv("PTAXSIM_DB_PATH")
+#)
 ```
 
-We can query the database table `agency_info` to search for `agency_num` key assigned the the taxing agencies of interest. This is what shows up when we search of agencies with the `CHICAGO` in the `agenc_name`.
+## Accounting for 2024 changes to agency fund reporting
+
+Before we query the relevant agency data, we first will need to check if any of the agencies of interest were impacted by the Clerk's 2024 updates to the data structure. We added new fields to the `agency_info` table to document these changes - we'll query this table and select only they agencies where the field `agency_change_24 = TRUE`. 
+
+```{r}
+# Query agency_info table for all agencies with the 2024 update
+agency_cw_24 <- DBI::dbGetQuery(
+  ptaxsim_db_conn,
+  "
+  SELECT *
+  FROM agency_info
+  WHERE agency_change_24 = TRUE
+  "
+) %>%
+  select(agency_num, agency_name, agency_num_24, agency_name_24)
+
+datatable(agency_cw_24)
+```
+
+From this table it appears that the updated agencies are specific funds for various municipalities - the Clerk had previously reported certain municipal fund levies as separate taxing agencies. This updated data structure begins in 2024 while years prior to 2024 remain the same, meaning users will need to account for this discrepancy if ever analyzing these taxing agencies data over time.
+
+To make this process easier, we have added new fields to the `agency_info` table in the PTAXSIM database which identify the agencies that have been folded into their parent agencies. `agency_num_24` and `agency_name_24` contain the agency info that the old "sub-agency" has been merged into. With this table we can create a crosswalk, as we did above, and call it `agency_cw_24`. Note that the user can still see details about these former agencies, now funds, by querying the `agency_fund` table.
+
+A quick search of the crosswalk shows that the `CITY OF CHICAGO LIBRARY FUND` was previously defined as its own agency and has now been folded into the `CITY OF CHICAGO`. This is in fact aligned with how the City of Chicago reports its property tax levy in its own budget [documentation](https://chicityclerk.s3.us-west-2.amazonaws.com/s3fs-public/O2023-0005291_Tax_Levy.pdf).
+
+
+
+We can query the database table `agency_info` to search for `agency_num` key assigned the the taxing agencies of interest. The table below displays all taxing agencies and TIFs with `CHICAGO` presenet in the `agency_name`.
 
 ```{r}
 chi_agency_nums <- DBI::dbGetQuery(
@@ -57,10 +84,10 @@ chi_agency_nums <- DBI::dbGetQuery(
   "
 )
 
-head(chi_agency_nums)
+datatable(chi_agency_nums)
 ```
 
-Note that there are several agencies that pop up when we search of agencies that contain `CHICAGO` in the name. Let's take a moment to examine the agencies `CITY OF CHICAGO` and `CITY OF CHICAGO LIBRARY FUND`, with agency fund numbers `030210000` and `030210002` respectively. While the `CITY OF CHICAGO LIBRARY FUND` has been treated as its own taxing agency in the Clerk's data, the City itself reports the amount levied for the purpose of funding the Chicago Public Library as a portion of its entire property tax levy.
+Let's take a moment to examine the agencies `CITY OF CHICAGO` and `CITY OF CHICAGO LIBRARY FUND`, with agency fund numbers `030210000` and `030210002` respectively. While the `CITY OF CHICAGO LIBRARY FUND` has been treated as its own taxing agency in the Clerk's data, the City itself reports the amount levied for the purpose of funding the Chicago Public Library as a portion of its entire property tax levy.
 
 We're also interested in the `BOARD OF EDUCATION` and `CHICAGO PARK DISTRICT` agencies, which have agency numbers `050200000` and `044060000` respectively. We'll query all fields from the `agency` table and filter by their `agency_num`. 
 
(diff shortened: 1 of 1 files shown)


vignettes/agencies.Rmd
```diff
@@ -0,0 +1,439 @@
+---
+title: "Keeping tabs on taxing agencies"
```
jeancochrane: **[Suggestion, optional]** I can see how this matches the tone of the "Tinkering with TIFs" doc, and it's a good title if we want to stick with that tone, but I personally prefer a more formal and descriptive tone for documentation like this. Maybe something like:


```suggestion
title: "Tracking taxing agency revenue over time"
```

vignettes/agencies.Rmd
```diff
@@ -0,0 +1,439 @@
...
+)
+```
+
+# Introduction
+
+One of the unique features of PTAXSIM is the database that accompanies the package - this DB contains the only publicly accessible, machine-readable data collected from the various Cook County property tax agencies going back to 2006. While the primary use of PTAXSIM is its functionality to calculate tax bills, its database holds a plethora of data that can be used to investigate Cook County's property tax system, including the behavior of the over 900 taxing agencies and 400 TIFs that collect revenue through property taxes.
+
+In this vignette, we'll demonstrate how to query data from the PTAXSIM SQLite database that can help us analyze taxing agencies' property tax revenue over time.
+
+Additionally, we'll show how to account for the 2024 changes to the Cook County Clerk's agency data structure which will be necessary for users to account for when conducting a time series analysis of certain taxing agencies.
```
jeancochrane: **[Nitpick, optional]** Slightly clearer wording:


```suggestion
Additionally, we'll show how to account for the 2024 changes to the Cook County Clerk's agency data structure, which is necessary when conducting a time series analysis of taxing agencies before and after 2024.
```

vignettes/agencies.Rmd
```diff
@@ -0,0 +1,439 @@
...
+
+# Chicago taxing agencies
+
+Using data from the PTAXSIM database, let's look at the levy history for three of the primary taxing agencies that appear on a Chicago property's tax bill: the `CITY OF CHICAGO`, `BOARD OF EDUCATION` (a.k.a. Chicago Public Schools), and `CHICAGO PARK DISTRICT`.
+
+First, load some useful libraries and instantiate a PTAXSIM DBI connection with the default name (`ptaxsim_db_conn`).
```
jeancochrane: **[Nitpick, optional]** I don't necessarily think `ptaxsim_db_conn` is the "default" name, though it is a name that we use often in our code. Not a huge deal either way, but I figure we can remove the adjective for extra clarity:


```suggestion
First, load some useful libraries and instantiate a PTAXSIM DBI connection with the variable name `ptaxsim_db_conn`.
```

vignettes/agencies.Rmd
```diff
@@ -0,0 +1,439 @@
...
+library(tidyr)
+```
+
+```{r, echo=FALSE}
+# This is needed to build the vignette using GitHub Actions
+ptaxsim_db_conn <- DBI::dbConnect(
+  RSQLite::SQLite(),
+  Sys.getenv("PTAXSIM_DB_PATH")
+)
+```
```
jeancochrane, line 46: **[Nitpick, required]** Since the block that assigns `ptaxsim_db_conn` is tagged with `echo=FALSE`, it won't get rendered in the report, meaning the report will not contain any example code for assigning a connection to `ptaxsim_db_conn`. I think we need to add a line to the `library()` block to render this example code, similar to how we do it in the [TIF vignette](https://github.com/ccao-data/ptaxsim/blob/eb1e3e8d4c7426656ea38b78d9d502b2f06286d2/vignettes/tifs.Rmd#L30-L50):

``````diff
```{r}
library(data.table)
library(dplyr)
library(DT)
library(here)
library(ggplot2)
library(ptaxsim)
library(tidyr)
+
+ ptaxsim_db_conn <- DBI::dbConnect(RSQLite::SQLite(), here("./ptaxsim.db"))
```
``````

File: vignettes/agencies.Rmd [hunk 5]
```diff
@@ -0,0 +1,439 @@
...
+```{r}
+chi_agencies <- DBI::dbGetQuery(
+  ptaxsim_db_conn,
+  "
+  SELECT DISTINCT *
+  FROM agency
+  WHERE agency_num = '030210000'
+  OR agency_num = '030210001'
+  OR agency_num = '050200000'
+  OR agency_num = '044060000'
```
jeancochrane: **[Nitpick, optional]** Slightly simpler SQL:


```suggestion
  WHERE agency_num IN ('030210000', '030210001', '050200000', '044060000')
```

vignettes/agencies.Rmd
```diff
@@ -0,0 +1,439 @@
...
+chi_agencies <- chi_agencies %>%
+  # Join the agency crosswalk to get the parent agency number for pre-2024 years
+  left_join(agency_cw_24, "agency_num") %>%
+  # for the agencies that did have an agency number change in 2024, replace the
+  # old agency_num with the new one
+  mutate(
+    agency_num =
+      ifelse(!is.na(agency_num_24),
+             agency_num_24,
+             agency_num)) %>%
```
jeancochrane: **[Suggestion, non-blocking]** It strikes me that the current data model for the 2024 agency number change requires the user to filter for agencies that have changed in 2024 (`agency_change_24`) in two different places: First in the SQLite query that constructs the crosswalk, and then again in this mutation that replaces `agency_num` with `agency_num_24` for agencies that have changed (which is technically not using `agency_change_24`, but is still using it indirectly through the conditional on `agency_num_24`). Since the data model is designed this way, users have to be careful to handle the 2024 change in both places, and doing that successfully requires them to have a strong mental model of how the data model works.

What if, instead of only populating `agency_num_24` for agencies where `agency_change_24 == TRUE` in the agency ingest script, we were to populate `agency_num_24` for _every_ agency, and just set it to the same value as `agency_num` for rows where `agency_change_24 == FALSE`? Then package users could remove the filter condition from `agency_cw_24`:

```r
agency_cw_24 <- DBI::dbGetQuery(
  ptaxsim_db_conn,
  "
  SELECT agency_num, agency_name, agency_num_24, agency_name_24
  FROM agency_info
  "
)
```

And we could simplify this mutation
```r
chi_agencies <- chi_agencies %>%
  left_join(agency_cw_24, "agency_num") %>%
  mutate(agency_num = agency_num_24) %>%
  # etc.
  ...
```

What do you think? I haven't thought this through very carefully, so I suspect there's a reason we designed it this way that I'm neglecting to think about.

vignettes/agencies.Rmd
```diff
@@ -0,0 +1,439 @@
...
+  group_by(year, agency_num) %>%
+  summarize(
+    total_final_levy = sum(total_final_levy),
+    total_ext = sum(total_ext)
+)
+```
+
+## Tracking Chicago, CPS, and CPKD levies
```
jeancochrane, line 125: **[Question, non-blocking]** I've never actually seen the Park District's agency name abbreviated as "CPKD", is that a standard abbreviation?

vignettes/agencies.Rmd
```diff
@@ -0,0 +1,439 @@
...
+
+In 2024, the Cook County Clerk's agency fund identifier keys were updated, now showing a greater level of detail than in prior years. You can learn more about this change and how we account for it in the PTAXSIM database by reading out 2024 update [changelog](https://ccao-data.github.io/ptaxsim/changelog). 
+
+To demonstrate working with the agency fund data, we will query the fund information for the `CITY OF CHICAGO` and the `CITY OF CHICAGO LIBRARY FUND`.
+
+The level of detail provided in`fund_name` varies across years and agencies which can make analysis or plotting the data tricky. To simplify `agency_fund` data further, we opted below to add broader categories to define funds.  In this case, we'll tag any fund with "A & B" (Annuities and Benefits) in the name as a pension fund. We'll label funds related to bond and interest payments, as well as note redemption, as "Bond Payments".
```
jeancochrane: **[Nitpick, optional]** Small spacing issues:


```suggestion
The level of detail provided in `fund_name` varies across years and agencies which can make analysis or plotting the data tricky. To simplify `agency_fund` data further, we opted below to add broader categories to define funds. In this case, we'll tag any fund with "A & B" (Annuities and Benefits) in the name as a pension fund. We'll label funds related to bond and interest payments, as well as note redemption, as "Bond Payments".
```

vignettes/agencies.Rmd
```diff
@@ -0,0 +1,439 @@
...
+
+```{r, echo=FALSE, out.width="100%"}
+chi_levy_plot_2
+```
+
+## Agency fund data updates and query demo
+
+The PTAXSIM database also contains information related to taxing agency's property tax funds so we can understand in greater detail what they intend to spend their property tax revenue on. This information can be found in the `agency_fund` and `agency_fund_info` tables. This data is not utilized by any of PTAXSIM's functions, but it is available to be queried an analyzed.
+
+In 2024, the Cook County Clerk's agency fund identifier keys were updated, now showing a greater level of detail than in prior years. You can learn more about this change and how we account for it in the PTAXSIM database by reading out 2024 update [changelog](https://ccao-data.github.io/ptaxsim/changelog).
```
jeancochrane: **[Nitpick, required]** The changelog link will actually be slightly different once it's deployed -- see [noctua's changelog](https://dyfanjones.github.io/noctua/news/) for an example path:


```suggestion
In 2024, the Cook County Clerk's agency fund identifier keys were updated, now showing a greater level of detail than in prior years. You can learn more about this change and how we account for it in the PTAXSIM database by reading out 2024 update [changelog](https://ccao-data.github.io/ptaxsim/news/). 
```

vignettes/agencies.Rmd
```diff
@@ -0,0 +1,439 @@
...
+  FROM agency
+  WHERE agency_num = '030210000'
+  OR agency_num = '030210001'
+  OR agency_num = '050200000'
+  OR agency_num = '044060000'
+  "
+)
+```
+
+Before we do anything, we need to fold the `CITY OF CHICAGO LIBRARY FUND` into the `CITY OF CHICAGO` levy total for all years prior to 2024. We can do this by joining `agency_cw_24` by `agency_num` to `chi_agencies` and replacing the old `agency_num` with `agency_num_24`, then grouping by agency and year and then summing the fields `total_final_levy` and `total_ext`. *Terminology note*: The **levy** is the amount of money a local government budgets for and requests to receive from property taxes. Many taxing agencies are subject to limits imposed by State law. The **extension** (called `final_ext` in the database) is the total, final amount a taxing body is allowed to receive which is calculated and confirmed by the Cook County Clerk.
```
jeancochrane: **[Suggestion, optional]** What do you think about turning this terminology note into a [hoverable footnote](https://bookdown.org/ewhayewon1123/seoyewon/footnotes-and-citations.html)?

Image


```suggestion
Before we do anything, we need to fold the `CITY OF CHICAGO LIBRARY FUND` into the `CITY OF CHICAGO` levy total for all years prior to 2024. We can do this by joining `agency_cw_24` by `agency_num` to `chi_agencies` and replacing the old `agency_num` with `agency_num_24`, then grouping by agency and year and then summing the fields `total_final_levy` and `total_ext`.[^1]

[^1]: *Terminology note*: The **levy** is the amount of money a local government budgets for and requests to receive from property taxes. Many taxing agencies are subject to limits imposed by State law. The **extension** (called `final_ext` in the database) is the total, final amount a taxing body is allowed to receive which is calculated and confirmed by the Cook County Clerk. 
```

vignettes/agencies.Rmd
```diff
@@ -0,0 +1,439 @@
...
+)
+```
+
+## Tracking Chicago, CPS, and CPKD levies
+
+Now that we have the correct total levies for the City of Chicago, Chicago Public Schools and Chicago Park District across all years, we can look at how those levies have changed from 2006 to 2024.
+
+CPS and CPKD are both subject to PTELL (Property Tax Extension Law Limit), which ensures certain taxing agencies do not increase their levies beyond the rate of inflation (with some exceptions\*). The City of Chicago is not by virtue of being a non-home rule agency. However, the City of Chicago imposes [its own limits](https://codelibrary.amlegal.com/codes/chicago/latest/chicago_il/0-0-0-2608573) which mirror PTELL's and prohibit a taxing agency from increasing its levy more than the rate of inflation or 5%, whichever is less.
+
+\*PTELL, as well as Chicago's property tax limitation ordinance, contain loopholes that allow taxing agencies to [increase their levies beyond the rate of inflation](https://civicfed.org/press/new-report-finds-illinois-property-tax-cap-law-not-working-intended). Let's see how Chicago's levy has fared compared to inflation since 2006.
```
jeancochrane: **[Suggestion, optional]** This seems like another good opportunity to use the built-in footnoting system:

Image


```suggestion
CPS and CPKD are both subject to PTELL (Property Tax Extension Law Limit), which ensures certain taxing agencies do not increase their levies beyond the rate of inflation (with some exceptions[^2]). The City of Chicago is not by virtue of being a non-home rule agency. However, the City of Chicago imposes [its own limits](https://codelibrary.amlegal.com/codes/chicago/latest/chicago_il/0-0-0-2608573) which mirror PTELL's and prohibit a taxing agency from increasing its levy more than the rate of inflation or 5%, whichever is less.

[^2]: PTELL, as well as Chicago's property tax limitation ordinance, contain loopholes that allow taxing agencies to [increase their levies beyond the rate of inflation](https://civicfed.org/press/new-report-finds-illinois-property-tax-cap-law-not-working-intended). Let's see how Chicago's levy has fared compared to inflation since 2006.
```

jeancochrane (changes requested):
Awesome work here! A couple small nitpicks and only one big thought related to the structure of the agency crosswalk data model. Otherwise, this is looking great.

At some point we should probably have Nicole review this for accuracy, though perhaps we should wait until it's in a closer to finished state? I leave it up to you to decide!

vignettes/agencies.Rmd
kyrasturgill: I copied this over from an existing vignette, so don't feel strongly about the language! I think your suggestion makes sense.

vignettes/agencies.Rmd
kyrasturgill: Oh, definitely, I must have accidentally deleted this line at some point. I'll add back in.

vignettes/agencies.Rmd
kyrasturgill: I actually had that in my head because Scott had referred to it as such in our correspondence ! If you think it would be confusing to the broader audience I can just spell it out here.

vignettes/agencies.Rmd
kyrasturgill: I see what you're saying - I think only problem is `agency_num_24` is NA for any agency that is not impacted by this change, so I think the mutate simplification would cause an error. But I can adjust the query.

jeancochrane: Ah yeah, to clarify, my suggested change would also require updating the way we construct the `agency_*_24` fields so that they are populated for all agencies, regardless of the value of `agency_change_24` -- for agencies where `agency_change_24` is false, our ingest script would set `agency_num` to the same value as `agency_num_24`. Does that make sense? If not, we can chat about it.

kyrasturgill: Ahhh gotcha so you're thinking it would be worth adjusting the agency ingest script?

jeancochrane: Punting this to #88!

jeancochrane (approved):
Ready to go!

jeancochrane:
Sorry to steal your thunder @kyrasturgill, but I'm going to merge this now so that I can get a jump on editing it in #89!

PR #86, opened 2026-04-22
Update README.md to reflect 2024 changes

### LINKED ISSUE #74: Make sure README is up to date with 2024 changes
We're making a number of changes to the code and the data model in order to support the 2024 data release because the underlying Clerk data changed so much this year. We should make sure the README is up to date with these changes. In particular, we need to document new and changed database tables.

Connects https://github.com/ccao-data/ptaxsim/issues/59.

kyrasturgill:
Necessary README updates to reflect the 2024 update. Not a ton to be changed here as I thought there would be. I opted to keep most things the same, barring where an update was required. Most substantial change is to the database diagrams, which I hope will render correctly!


README.Rmd
```diff
@@ -217,10 +217,10 @@ multiple_years_plot <- ggplot(data = multiple_years_summ) +
   scale_y_continuous(
     name = "Total Tax Amount",
     labels = scales::dollar,
-    expand = c(0, 0),
+    expand = expansion(mult = c(0, .05)),
```
kyrasturgill, line 220: With the dramatic 2024 increase I thought the axis spacing was making it look as though it was cut off, so I added some more buffer here.

README.Rmd
```diff
@@ -264,7 +266,8 @@ The PTAXSIM backend database contains cleaned data from the Cook County Clerk, T
 
 ## Notes and caveats
 
-- Currently, the per-district tax calculations for properties in the Red-Purple Modernization (RPM) TIF are slightly flawed. However, the total tax bill per PIN is still accurate. See issue [#4](https://github.com/ccao-data/ptaxsim/issues/4) for more information or issue [#56](https://github.com/ccao-data/ptaxsim/issues/56).
```
kyrasturgill, line 267: I believe this has all been resolved with our transit TIF issue update right?

inst/mermaid/er-diagram-big.mmd
```diff
@@ -56,12 +48,14 @@ erDiagram
...
     double  prelim_rate
     bigint  ptell_reduced_levy
-    boolean ptell_reduced_ind
     bigint  final_levy
     double  final_rate
   }
 
   agency_fund_info {
+    varchar agency_num         PK
+    varchar fund_type_num      PK
```
jeancochrane: **[Nitpick, optional]** Since `fund_type_num` is derived from `fund_num`, and `fund_num` is already part of the primary key, I think this field does not need to be part of the primary key, right?


```suggestion
    varchar fund_type_num
```

README.Rmd
jeancochrane: **[Question, non-blocking]** It's a bit hard to tell since the Y-axes are not aligned, but it looks as if the increment value is now larger between 2016 and 2023 than it was in the previous version of the chart, even setting aside the big jump in 2024 (which is also pretty shocking, so much so that we should fact check it). Do you see that too? If so, any idea why might be the case?

README.Rmd
jeancochrane: **[Question, non-blocking]** Hmm, we didn't remove this note when we fixed the `tax_bill(simplify = TRUE)` logic to account for transit TIFs in https://github.com/ccao-data/ptaxsim/pull/58; instead, we added onto it. Now that I have a better understanding of how everything in this repo fits together, I might phrase this note slightly differently:

```Rmd
- The per-district tax calculation using `tax_bill(simplify = TRUE)` for properties in transit TIFs do not match the amounts that the Treasurer reports on their tax bills. We believe the amounts we report are correct, however. See issues [#4](https://github.com/ccao-data/ptaxsim/issues/4) and [#56](https://github.com/ccao-data/ptaxsim/issues/56) for more information, as well as PR [#58](https://github.com/ccao-data/ptaxsim/pull/58).
```

@erhla If you have a minute to weigh in here, it would be helpful! If not, we'll make our best guess.

inst/mermaid/er-diagram-big.mmd
```diff
@@ -167,6 +162,23 @@ erDiagram
...
+    int     pin_increment_eav
+    double  pin_distribution_pct
+    double  transit_tif_to_cps
+    double  transit_tif_to_tif
+    double  transit_tif_to_dist
+    boolean is_transit_tif
+  }
+
   eq_factor ||--|{ pin : "applies to"
   pin ||--|{ tax_code : "within"
```
jeancochrane, line 183: **[Suggestion, optional]** What do you think about adding a "may have" relationship between `pin` and `pin_tif_distribution`? Or another type of relationship, if "may have" is not appropriate (e.g. "has")?


```suggestion
  pin ||--|{ tax_code : "within"
  pin ||--o| pin_tif_distribution : "may have"
```

inst/mermaid/er-diagram-small.mmd
```diff
@@ -90,6 +90,17 @@ erDiagram
...
+    varchar agency_num         PK
+    varchar tax_code_num
+    double  pin_distribution_pct
+    double  transit_tif_to_cps
+    double  transit_tif_to_tif
+    double  transit_tif_to_dist
+  }
+
   eq_factor ||--|{ pin : "applies to"
   pin ||--|{ tax_code : "within"
```
jeancochrane, line 103: **[Suggestion, non-blocking]** A quick note that if we modify relationships in the large diagram above per my suggestion, we should modify them here, too.

inst/mermaid/er-diagram-small.mmd
```diff
@@ -90,6 +90,17 @@ erDiagram
...
 
+  pin_tif_distribution {
+    int     year               PK
+    varchar pin                PK
+    varchar agency_num         PK
+    varchar tax_code_num
+    double  pin_distribution_pct
+    double  transit_tif_to_cps
+    double  transit_tif_to_tif
+    double  transit_tif_to_dist
```
jeancochrane: **[Nitpick, optional]** I wonder if these `transit_tif_to_*` fields are even worth including in this small version of the diagram? They're definitely important, but I also think we could cut them for further simplicity.

inst/mermaid/er-diagram-small.mmd
jeancochrane, line 30: **[Nitpick, blocking]** I think we may need to update this table to add `agency_name` and make it part of the primary key, right?

jeancochrane (changes requested):
Nice work here! I'm glad this is so simple. There's only one issue with the small diagram that I think is blocking; otherwise, consider everything else below to be suggestions.

README.Rmd
kyrasturgill: I see that too when looking at the diff - this .png was generated when I knitted the file but when I am running that code chunk alone it appears correct and aligned with the old viz. Weird, not sure why that is happening!
Here is image I get when I just ran it:
image

And as for the dramatic spike, this PIN's AV spiked in 2024, going from 7000 to 12500. Perhaps we should select a PIN with more stable AV?

README.Rmd
kyrasturgill: Ah right - I am surprised we still described the calculations as flawed because I thought we had corrected the issues that are linked to. But agreed, there is still the discrepancy  between the per-district tax calculations and what is printed on the tax bills so we should keep a note in here. I like what you drafted here!

inst/mermaid/er-diagram-small.mmd
kyrasturgill: I was debating this as well, I am fine with removing

inst/mermaid/er-diagram-small.mmd
kyrasturgill: Ah good catch - yes for agency_num.

README.Rmd
jeancochrane: > this .png was generated when I knitted the file but when I am running that code chunk alone it appears correct and aligned with the old viz. Weird, not sure why that is happening!

That's so strange! Let me know if you want a second eye debugging it. I get the correct and more aligned version when I run it and knit it locally too, although I have to tweak the two `DBI::dbConnect()`  paths to point to the correct database -- I wonder if it's possible that you have two different versions of the database, one of which gets loaded instead of the other when knitting vs. running locally? Very low confidence in that diagnosis, it's just an idea as to the kind of thing that might be going wrong. In any case, I think we should make sure we have the correct / aligned version of the chart before we merge.

> And as for the dramatic spike, this PIN's AV spiked in 2024, going from 7000 to 12500. Perhaps we should select a PIN with more stable AV?

Got it, that makes sense! I don't mind keeping it as is, it's kind of interesting to see such a dramatic spike in 2024.

kyrasturgill:
@jeancochrane I believe I implemented all requested edits - only thing I'm unclear on is why the viz is off when the doc is rendered.

data-raw/create_db.sql
```diff
@@ -83,12 +83,12 @@ CREATE INDEX ix_agency_fund_fund_num ON agency_fund(fund_num);
 /** agency_fund_info **/
 CREATE TABLE agency_fund_info (
     agency_num              varchar(9)                                 NOT NULL,
-    fund_type_num           varchar(3)                                 NOT NULL,
+    fund_type_num           varchar(6)                                 NOT NULL,
```
jeancochrane: **[Question, non-blocking]** Was this change intentional?

jeancochrane (approved):
Looks great, barring one small change to the `fund_type_num` definition in `create_db.sql` that I can't quite explain! Let's plan to pair on Tuesday to see if we can figure out why the knitted chart looks different from the one that renders locally. Once we get that resolved, this is good to merge.

data-raw/create_db.sql
kyrasturgill: I'm losing it, I must have ready this as `fund_num` and thought I'd made a mistake. I will revert this change.

README.Rmd
kyrasturgill: I think I have addressed the problem. When knitting the doc it was loading the existing version of the `ptaxsim` package rather than the dev version which meant TIF amounts were not being calculated correctly. I knit the doc with it loading the dev version and pushed the updated viz .png and .md files but I'm not 100% sure that was the right way to do it

data-raw/create_db.sql
```diff
@@ -88,7 +88,7 @@ CREATE TABLE agency_fund_info (
     fund_num                varchar(6)                                 NOT NULL,
     fund_name               varchar                                    NOT NULL,
     capped_ind              boolean                                    NOT NULL,
-    PRIMARY KEY (agency_num, fund_type_num, fund_num)
+    PRIMARY KEY (agency_num, fund_num)
```
jeancochrane, line 91: **[Question, non-blocking]** Just to clarify: I believe we will need to do a new database export to reflect this change, correct? It's not a big deal -- the most important advantages of primary keys are uniqueness constraints and indexes, neither of which is particularly important for this use-case -- but I just want to make sure my understanding is correct.

jeancochrane (approved):
Thanks for getting to the bottom of the chart rendering issue!

data-raw/create_db.sql
kyrasturgill: Yep that's correct! Which we will ultimately need to do now anyway to incorporate the new agency crosswalks.

PR #89, opened 2026-04-29
Refactor agency crosswalk and add agency fund crosswalk

### LINKED ISSUE #88: Refactor agency crosswalk data model to better reflect underlying data change
## Background and problem statement

Now that I have finally internalized the 2024 change to agencies that has inspired the agency crosswalk (https://github.com/ccao-data/ptaxsim/pull/65) and its associated vignette (https://github.com/ccao-data/ptaxsim/pull/84) -- thanks in large part to @kyrasturgill's patience and wisdom -- I have some thoughts regarding how we might refactor the crosswalk data model to make it more intuitive to use. We've discussed some of these ideas in person and in Teams chats, but I want to pull the discussion out into an issue so that we can preserve it for posterity.

As a reminder, here is how we are currently implementing the "agency crosswalk" in 2024:

* New columns on the `agency_info` table
    * `agency_info.agency_num_24` (optional string): New agency number in 2024, or null if the agency number has not changed
    * `agency_info.agency_name_24` (optional string): New agency name in 2024, or null if the agency name has not changed
    * `agency_info.agency_change_24` (required boolean): True if the agency changed in 2024, and false otherwise

To me, the issue is that this data model is framed around the notion that **the Clerk has changed some agency numbers in 2024** -- hence column names like `agency_num_24` (in contrast to `agency_num`) and `agency_change_24`, which lead the user to understand the change to be a change of agency number. However, based on recent conversations like https://github.com/ccao-data/ptaxsim/pull/85#pullrequestreview-4171660806, it's become apparent to me that it would be more accurate to frame the issue by saying that **the Clerk has _changed some agencies to funds_ in 2024**: The Clerk reported some funds (like the City of Chicago library funds) as independent agencies prior to 2024, but now they report these funds as funds underneath a parent agency (in this case, the City of Chicago). The old independent agencies still exist in the 2024 agency report, but they all have $0 levies.

If this updated framing is indeed more accurate, then I think the current way we structure the "agency crosswalk" is slightly misleading, since it elides the substance of the change, implying that the crux of the change is that some agencies have new numbers, when in reality those agencies have become subsumed as funds under different agencies.

You could argue that this is not a particularly important concern, since we could avoid misunderstandings with more documentation. However, I think the current design also has a couple of concrete disadvantages:

- **Disadvantage 1: The current data model ignores 2024 changes to fund numbers**. In the process of moving some funds from an independent agency to a parent agency, the clerk has also changed some fund numbers. For example, here is the Note Redemption fund for CPL in 2023:

| year | agency_num | fund_num | agency_name | fund_name | levy |
| --- | --- | --- | --- | --- | --- |
| 2023 | 030210001 | 259000 | CITY OF CHICAGO LIBRARY FUND  | LIBRARY NOTE REDEMPTION | 114582000 |

Here is the same fund in 2024:

| year | agency_num | fund_num | agency_name | fund_name | levy |
| --- | --- | --- | --- | --- | --- |
| 2024 | 030210000 | 259001 | CITY OF CHICAGO | NOTE REDEMPTION LIBRARY FUND | 117145000 |

Note that the both the agency number and fund number have changed. If the user doesn't care about specific funds, this is probably fine, since they will just roll all of the funds up to the agency level anyway. But if the user wants to track specific funds -- say, just CPL funds -- they will have to do some digging to figure out the new fund numbers and hardcode those fund numbers into their analysis.

- **Disadvantage 2: The current data model will not be adaptable if these types of changes occur again in the future**. The fields are currently tightly coupled to the expectation that this is a one-time change in 2024. If a similar change were to take place next year, we would need a new set of fields like `agency_info.agency_change_25`. It is our expectation (and hope) that this kind of change won't happen again, but if it does, our data model won't handle it well.

## Proposed solutions

Here are a few options for how I think we might improve the agency crosswalk data model to resolve the issues I raised above.

### Option 1: Dedicated crosswalk tables

This option splits out the `agency_*_24` columns into dedicated tables, adds a more flexible `year` column, and adds support for converting fund numbers in addition to agency numbers. Read on for descriptions of the proposed tables.

#### Agency crosswalk

- New table `agency_crosswalk` that would replace the `agency_info.agency_*_24` fields that we currently maintain
    - Fields contained in this table: 
        - `year` (required integer): Year that `agency_num` changed
        - `agency_num` (required string): The old agency number, prior to the change
        - `agency_num_final` (required string): The new version of the agency number following the change
    - Table would have one row for each agency that has changed at some point in time, with `year` set to the year that the agency changed

Here's an example of the rows that would exist in `agency_crosswalk` to support the change to the CPL fund:

| year | agency_num | agency_num_final |
| --- | --- | --- |
| 2024 | 030210001 | 030210000 |

Here's how a user could use this table:

```R
agency_cw = DBI::dbGetQuery(ptaxsim_db_conn, "SELECT * FROM agency_crosswalk")

agencies %>%
  left_join(agency_cw, by = c("agency_num")) %>%
  mutate(agency_num = coalesce(agency_num_final, agency_num))
```

Note that the user doesn't actually need to use the `year` column, because we are assuming that agency/fund transitions are one-way. If that assumption ever gets violated in the future, we would need to reconsider our approach, because even including the `year` column in the join wouldn't be enough to resolve the potential ambiguity. (More on that in the pros/cons section below.)

#### Fund crosswalk

- New table `agency_fund_crosswalk`
    - `year` (required integer): Year that that `agency_num` and/or `fund_num` changed
    - `agency_num` (required string): The old agency number, prior to the change
    - `agency_num_final` (required string): The new agency number, after the change
    - `fund_num` (required string): The old fund number, prior to the change
    - `fund_num_final` (required string): The new fund number, after the change

Here's what `agency_fund_crosswalk` would like like for the CPL Note Redemption fund:

| year | agency_num | agency_num_final | fund_num | fund_num_final |
| --- | --- | --- | --- | --- |
| 2024 | 030210001 | 030210000 | 259000 | 259001 |

Here's how a user could use this table:

```R
agency_fund_cw = DBI::dbGetQuery(ptaxsim_db_conn, "SELECT * FROM agency_fund_crosswalk")

agency_funds %>%
  left_join(agency_fund_cw, by = c("agency_num", "fund_num") ) %>%
  mutate(
    agency_num = coalesce(agency_num_final, agency_num),
    fund_num = coalesce(fund_num_final, fund_num)
)
```

#### Pros and cons of this approach

* **Pros** of this approach:
    * Meets three requirements motivating this issue:
        1. More closely matches the underlying data change
        2. Supports tracking changing funds over time using fund numbers
        3. Easy for our team to update over time, since it's just keyed by year and agency number
    * Very similar to our current design, but adds easy support for new years in the future
    * Easier for end users to understand and use than our current approach
* **Cons** of this approach:
    * Medium-sized refactor that would take some time to implement
    * Introduces two new tables that we have to maintain
    * If a fund ever changed _back_ to being an independent agency, this data model would break down, and we would need to replace it with a more sophisticated one (like option 2 below); this is due to the fact that if two agencies get consolidated at a certain point in time, and then get disaggregated at a future point in time, this crosswalk structure won't tell us _which_ of those two agencies got disaggregated
 
### Option 2: Dedicated crosswalk tables, filled for all years

This option modifies the tables proposed in [option 1 above](#option-1) to fill all years prior to a change in agency/fund number. This small tweak makes `year` into a necessary column for joins, and makes the solution robust to the possibility that the Clerk could change some funds back to independent agencies in the future.

#### Agency crosswalk

- New table `agency_crosswalk` that would replace the `agency_info.agency_*_24` fields that we currently maintain
    - Columns contained in this table: 
        - `year` (required integer): Year that the `agency_num` was active
        - `agency_num` (required string): Agency number in this current year
        - `agency_num_final` (required string): The final, canonical agency number for this agency that is guaranteed to be stable across years
            - When generating `agency_crosswalk` each year, we would set this value to the `agency_num` value for the agency in the most recent year (2024, for the current update) 
    - Table would have one row for each agency that changed _in each year leading up to the change_

Here's how `agency_crosswalk` could support the change to the CPL fund. Rows start with the year 2006 since that's the first year that PTAXISM has data for this fund:

| year | agency_num | agency_num_final |
| --- | --- | --- |
| 2006 | 030210001 | 030210000 |
| 2007 | 030210001 | 030210000 |
| 2008 | 030210001 | 030210000 |
| 2009 | 030210001 | 030210000 |
| 2010 | 030210001 | 030210000 |
| 2011 | 030210001 | 030210000 |
| 2012 | 030210001 | 030210000 |
| 2013 | 030210001 | 030210000 |
| 2014 | 030210001 | 030210000 |
| 2015 | 030210001 | 030210000 |
| 2016 | 030210001 | 030210000 |
| 2017 | 030210001 | 030210000 |
| 2018 | 030210001 | 030210000 |
| 2019 | 030210001 | 030210000 |
| 2020 | 030210001 | 030210000 |
| 2021 | 030210001 | 030210000 |
| 2022 | 030210001 | 030210000 |
| 2023 | 030210001 | 030210000 |
| 2024 | 030210000 | 030210000 |

Here's how a user could use this table:

```R
agency_cw = DBI::dbGetQuery(ptaxsim_db_conn, "SELECT * FROM agency_crosswalk")

agencies %>%
  left_join(agency_cw, by = c("year", "agency_num")) %>%
  mutate(agency_num = coalesce(agency_num_final, agency_num))
```

Note that, in contrast to option 1 above, `year` is now part of the join key. This gives us greater control over the changes that happen year-over-year, which could be useful in particular if the Clerk changes back to reporting certain funds as independent agencies.

#### Fund crosswalk

- New table `agency_fund_crosswalk`
    - Columns contained in this table: 
        - `year` (required integer): Year that the `agency_num` and `fund_num` were active
        - `agency_num` (required string): Agency number in this year
        - `agency_num_final` (required string): Canonical agency number that is stable across years (i.e. number for the most recent year)
        - `fund_num` (required string): Fund number in this year
        - `fund_num_final` (required string): Canonical fund number that is stable across years (i.e. number for the most recent year)
    - Table would have one row for each fund that changed in each year leading up to the change

Here's what `agency_fund_crosswalk` would like like for the CPL Note Redemption fund:

| year | agency_num | agency_num_final | fund_num | fund_num_final |
| --- | --- | --- | --- | --- |
| 2006 | 030210001 | 030210000 | 259000 | 259001 |
| 2007 | 030210001 | 030210000 | 259000 | 259001 |
| 2008 | 030210001 | 030210000 | 259000 | 259001 |
| 2009 | 030210001 | 030210000 | 259000 | 259001 |
| 2010 | 030210001 | 030210000 | 259000 | 259001 |
| 2011 | 030210001 | 030210000 | 259000 | 259001 |
| 2012 | 030210001 | 030210000 | 259000 | 259001 |
| 2013 | 030210001 | 030210000 | 259000 | 259001 |
| 2014 | 030210001 | 030210000 | 259000 | 259001 |
| 2015 | 030210001 | 030210000 | 259000 | 259001 |
| 2016 | 030210001 | 030210000 | 259000 | 259001 |
| 2017 | 030210001 | 030210000 | 259000 | 259001 |
| 2018 | 030210001 | 030210000 | 259000 | 259001 |
| 2019 | 030210001 | 030210000 | 259000 | 259001 |
| 2020 | 030210001 | 030210000 | 259000 | 259001 |
| 2021 | 030210001 | 030210000 | 259000 | 259001 |
| 2022 | 030210001 | 030210000 | 259000 | 259001 |
| 2023 | 030210001 | 030210000 | 259000 | 259001 |
| 2024 | 030210000 | 030210000 | 259001 | 259001 |

Here's how a user could use this table:

```R
agency_fund_cw = DBI::dbGetQuery(ptaxsim_db_conn, "SELECT * FROM agency_fund_crosswalk")

agency_funds %>%
  left_join(agency_fund_cw, by = c("year", "agency_num", "fund_num") ) %>%
  mutate(
    agency_num = coalesce(agency_num_final, agency_num),
    fund_num = coalesce(fund_num_final, fund_num)
)
```

#### Pros and cons of this approach

* Pros:
    * Meets the three requirements motivating this issue
    * Future-proof to the universe of possible changes that the Clerk could make to funds and agencies in the future
    * Very similar to how `tif_crosswalk` works, which is nice from the perspective of standardization
* Cons:
    * Hard to immediately tell which data changed when based on a cursory glance at either crosswalk; you can derive the change by looking for the year when `agency_num == agency_num_final` or `fund_num == fund_num_final`, but that's not very intuitive
    * Needing to populate rows for all years prior to a change leads to a lot of extraneous rows in the table that don't contain much information
    * Clerk _probably_ won't switch back to reporting funds as independent agencies in the future, so this solution may represent premature optimization

### Option 3: Keep the current approach

* Pros:
    * Requires no extra work, so it's fast to ship
    * Implements the fewest possible changes to the data model
* Cons:
    * Doesn't meet any of the three requirements motivating this issue
    * May be challenging for end users to parse and understand

### Summary and recommendation

Here's how I would summarize the costs and benefits of each option:

* **Option 1** (Dedicated crosswalk tables): Medium refactor, medium benefit to the user
* **Option 2** (Dedicated crosswalk tables that can be joined by `year` and filled automatically): Big refactor, big benefit to the user
* **Option 3** (Keep the current approach): Minimal refactor, minimal user benefit
    
Based on these tradeoffs, my suggestion is that we implement **option 1**, though I'm open to pushback or alternative ideas.

---

Curious what you think about this idea @kyrasturgill! Also happy to talk more on a call next week.

jeancochrane:
This PR refactors the agency crosswalk data model to move crosswalk columns out of the `agency_info` table and into two dedicated tables `agency_crosswalk` and `agency_fund_crosswalk`. There are two main advantages to this change:

1. Both tables include a `year_of_change` column recording the year of the change, which will allow us to easily update the crosswalk in the future if agency or fund numbers change again
2. The `agency_fund_crosswalk` table allows for resolving individual funds over time across the change in the same way that the agency crosswalk allows for resolving agencies over time

Closes https://github.com/ccao-data/ptaxsim/issues/88. See that issue for more details about this design and the reasons why we chose it.

#### Agency crosswalk

I tested to confirm that the contents of the new agency crosswalk exactly match the contents of the existing agency crosswalk:

```R
agency_cw_old <- DBI::dbGetQuery(
  db_conn_old,
  "select agency_num, agency_num_24 AS agency_num_final
  from agency_info where agency_change_24"
) %>%
  rename_with(~ paste0(., "_old"))

agency_cw_new <- DBI::dbGetQuery(
  db_conn_new,
  "select agency_num, agency_num_final
  from agency_crosswalk"
) %>%
  rename_with(~ paste0(., "_new"))

agency_cw_new %>%
  full_join(
    agency_cw_old,
    by = c("agency_num_new" = "agency_num_old"),
    keep = TRUE
  ) %>%
  mutate(
    chk_agency_num_in_old_not_new = is.na(agency_num_new),
    chk_agency_num_in_new_not_old = is.na(agency_num_old),
    chk_agency_num_final_mismatch = (
      !is.na(agency_num_new) &
        !is.na(agency_num_old) &
        agency_num_new != agency_num_old
    )
  ) %>%
  filter(if_any(starts_with("chk_"), ~ .x))
```

#### Agency fund crosswalk

QCing the agency fund crosswalk was harder because no source data exists for it and there isn't an existing version we can compare it to. I ended up double-checking each of the new fund numbers against the 2024 agency report Detail tab to confirm that they looked correct.

This surfaced one issue:

- A few funds exist in theory but not in practice -- that is to say, they had an entry in 2023, and based on our fund numbering logic they should exist in 2024, but they do not. Based on my investigation, I believe these are all funds that had $0 levies in 2023 and so got removed from the 2024 report, but their theoretical fund numbers are still correct.
    - These funds include:
        - 206001 (library building notes -- only existed in Elmwood Park pre-2024, but had a $0 levy)
        - 003001 (library bonds -- missing for many _but not all_ agencies in the crosswalk)
        - 003002 (general assistance bonds -- only existed for Orland in 2023, but had a $0 levy)
        - 003005 (public health bonds -- only existed for Stickney in 2023)
        - 027001 (library purchase agreement -- only existed for Markham in 2023 with $0 levy)

Otherwise, the fund numbering logic appears correct to me based on my QC.

#### Vignettes

I ran the vignettes locally to confirm that they render as expected. (Doing so requires downloading the new pre-release database version.)


data-raw/agency/agency.R
```diff
@@ -576,59 +582,86 @@ agency_info <- agency_info %>%
...
+arrow::write_parquet(
+  x = agency %>% select(-agency_name),
+  sink = remote_path_agency,
+  compression = "zstd"
+)
+arrow::write_parquet(
+  x = agency_info,
+  sink = remote_path_agency_info,
+  compression = "zstd"
+)
```
jeancochrane, line 595: These lines are not new, the diff is just confusing because some of the agency crosswalk code is extracted from the old `agency_info` code and git is confused about it.

data-raw/agency/agency.R
```diff
@@ -576,59 +582,86 @@ agency_info <- agency_info %>%
...
+      # handle them separately
+      fund_type_num == "408" ~ fund_num,
+      minor_type == "LIBRARY" ~ paste0(fund_type_num, "001"),
+      minor_type == "GEN ASST" ~ paste0(fund_type_num, "002"),
+      minor_type == "INFRA" ~ paste0(fund_type_num, "003"),
+      minor_type == "HEALTH" &
+        str_detect(agency_name, "MENTAL") ~ paste0(fund_type_num, "004"),
+      minor_type == "HEALTH" &
+        str_detect(agency_name, "PUBLIC") ~ paste0(fund_type_num, "005")
+    )
```
jeancochrane, line 660: These rules are not documented as far as I know, but I determined them through trial and error. My QC of all of the resulting fund numbers confirmed that they are correct.

data-raw/create_db.R
```diff
@@ -36,17 +36,18 @@ db_send_queries <- function(conn, sql) {
 # Set the database version. This gets incremented manually whenever the database
 # changes. This is checked against Config/Requires_DB_Version in the DESCRIPTION
 # file via check_db_version(). Schema is:
-# "MAX_YEAR_OF_DATA.MAJOR_VERSION.MINOR_VERSION-PRE_RELEASE_VERSION"
```
jeancochrane, line 39: I only just realized that this line is documenting the format for `db_version`, but `db_pre_release_version` is separate. I moved the pre-release component of this documentation to a separate comment on `db_pre_release_version`.

data-raw/create_db.R
```diff
@@ -139,10 +141,15 @@ datasets <- c("pin", "tax_code")
...
-    # the first file it finds in the dataset
+    # Starting in 2024, the data source for the `pin` table has changed.
+    # The current package maintainers do not have access to the old data source,
+    # which was a snapshot mirror of a mainframe system that has been
+    # decommissioned. To facilitate future updates, we copied over pre-2024
+    # `pin` files without edits. These legacy files are missing some columns
+    # that we added in 2024, so we need to unify the schemas across files, since
+    # otherwise arrow will take the schema from the first file it finds in the
+    # dataset. In the future, we could also consider editing the old files to
+    # add empty values for the new columns
```
jeancochrane, line 152: The prior iteration of this comment was slightly misleading, so I updated it for accuracy and clarity.

data-raw/create_db.sql
```diff
@@ -46,16 +46,22 @@ CREATE TABLE agency_info (
...
 
 
+/** agency_crosswalk **/
+CREATE TABLE agency_crosswalk (
+    year                    int                                        NOT NULL,
+    agency_num              varchar(9)                                 NOT NULL,
+    agency_num_final        varchar(9)                                 NOT NULL,
+    PRIMARY KEY (agency_num)
+) WITHOUT ROWID;
+
```
jeancochrane, line 63: No indexes on either of these crosswalks because the primary keys add automatic indexes. The tables are also pretty small so the indexes shouldn't provide much performance improvement anyway.

inst/mermaid/er-diagram-big.mmd
```diff
@@ -32,9 +32,12 @@ erDiagram
     varchar agency_name_original
```
jeancochrane, line 32: I find it helpful to use this site for live previews of Mermaid diagrams: https://mermaid.ai/live/edit

DESCRIPTION
```diff
@@ -3,7 +3,7 @@ Type: Package
 Title: Calculate Cook County Property Tax Bills and Simulate Scenarios
 Version: 1.1.0
 Authors@R: c(
-    person(given = "Kyra", family = "Sturgill", email = "kyra.sturgill@cookcountyil.gov", role = c("aut", "cre")),
+    person(given = "Kyra", family = "Sturgill", email = "Assessor.Data@cookcountyil.gov", role = c("aut", "cre")),
```
jeancochrane, line 6: Changing this from your personal email to the shared team email for the sake of privacy and durability. I don't feel strongly about it though, if you'd rather have your personal email listed we can keep it that way.

README.Rmd
```diff
@@ -240,20 +240,22 @@ The PTAXSIM backend database contains cleaned data from the Cook County Clerk, T
...
-| tif_crosswalk    | Clerk             | Manually created from TIF summary and distribution reports               | [data-raw/tif/tif.R](data-raw/tif/tif.R)                         | Fix for data issue identified in #39                |
-| tif_distribution | Clerk             | [TIF Reports - Tax Increment Agency Distribution Reports](https://www.cookcountyclerkil.gov/property-taxes/tifs-tax-increment-financing/tif-reports) | [data-raw/tif/tif.R](data-raw/tif/tif.R)                         | TIF EAV, frozen EAV, and distribution percentage by tax code      |
-| pin_tif_distribution | Clerk | [TIF Reports - Tax Increment Agency Distribution Reports](https://www.cookcountyclerkil.gov/property-taxes/tifs-tax-increment-financing/tif-reports) | [data-raw/tif/tif.R](data-raw/tif/tif.R) | TIF EAV, frozen EAV, and distribution percentage by PIN |
+| Table Name            | Source Agency     | Source Link                                                                                                                                          | Ingest Script                                                    | Contains                                                          |
```
jeancochrane, line 243: The main diff here is adding rows for `agency_crosswalk` and `agency_fund_crosswalk`. The rest of it is just whitespace changes to reformat the table so that it fits `agency_fund_crosswalk`, which is now the longest table name in the table.

vignettes/agencies.Rmd
```diff
@@ -47,74 +47,126 @@ ptaxsim_db_conn <- DBI::dbConnect(
...
+
+The following types of funds were affected by this change:
+
+- Library funds
+- General assistance funds
+- Infrastructure funds (road and bridge)
+- Mental health and public health funds
+- Levy adjustments
+
+Most tax codes contain at least one of these types of agencies.
```
jeancochrane, line 60: I added some background here to give the reader context regarding the 2024 change.

vignettes/agencies.Rmd
```diff
@@ -47,74 +47,126 @@ ptaxsim_db_conn <- DBI::dbConnect(
 
 ## Accounting for 2024 changes to agency fund reporting
```
jeancochrane, line 49: I took a second pass at this doc and made a number of changes to the code/prose in addition to swapping out the crosswalk format. I hope that's OK! Down to make further edits if you disagree with any of my choices here.

vignettes/agencies.Rmd
```diff
@@ -295,11 +347,9 @@ chi_levy_plot_2 <- ggplot(plot_2_df, aes(year, total_ext, group = agency_num)) +
 chi_levy_plot_2
 ```
 
-## Agency fund data updates and query demo
+## Tracking specific fund revenue over time
```
jeancochrane, line 350: This section header feels clearer to me, though I'm open to pushback.

data-raw/agency/agency.R
```diff
@@ -576,59 +582,91 @@ agency_info <- agency_info %>%
...
+  # Account for error in Clerk's report which lists the Skokie Library fund
+  # levy in the legacy fund line. Because of this, it continues to follow the
+  # pre-2024 pattern for library fund reporting, so we need to remove it from
+  # the crosswalk
+  filter(!(agency_num == "031170001" & agency_num_final == "031170000"))
+
+
+# agency_fund_crosswalk --------------------------------------------------------
+
+changed_funds <- agency_fund_info %>%
```
kyrasturgill, line 628: **[question, non-blocking]** Just checking my understanding is correct - `changed_funds` technically refers to the funds that previously were under the agencies that starting in 2024 have become funds? Not funds that are "new" (i.e. don't end in "000")?

data-raw/agency/agency.R
```diff
@@ -576,59 +582,91 @@ agency_info <- agency_info %>%
...
-    agency_num_24,
-    agency_name_24,
-    agency_change_24
+  left_join(agency_crosswalk, by = "agency_num") %>%
+  filter(!is.na(agency_num_final)) %>%
+  mutate(
+    fund_num_final = case_when(
+      # Levy adjustments (408) have the same fund numbers across all years, so
+      # handle them separately
+      fund_type_num == "408" ~ fund_num,
```
kyrasturgill, line 647: **[question, non-blocking]:** Technically there's nothing "new" about this fund right? The purpose of this being in the crosswalk is for a user who may want to track specifically total recapture pre- and post- 2024 for one of the agencies that had one of these changed funds?

data-raw/agency/agency.R
kyrasturgill: This makes sense and is a smart workaround!

data-raw/create_db.R
kyrasturgill: Helpful!

data-raw/agency/agency.R
```diff
@@ -576,59 +582,91 @@ agency_info <- agency_info %>%
...
   )
+}
+if (any(is.na(changed_funds$fund_num_final))) {
+  stop(
+    "Could not parse final fund number for some funds in agency_fund_crosswalk"
+  )
+}
+
+agency_fund_crosswalk <- changed_funds %>%
+  mutate(year = "2024") %>%
```
kyrasturgill: **[question]**: Is the intention here to include for year to denote for which tax year the `agency_num` change was introduced? I just am wondering if that could cause some friction when joining the crosswalk because it would not be useful/correct to join by `agency_num` and `year`. Does that make sense?

kyrasturgill: I now see your explanation of what `year` means in the vignette which pretty perfectly answers this question! I can see why it makes sense to keep it.

vignettes/agencies.Rmd
```diff
@@ -47,74 +48,126 @@ ptaxsim_db_conn <- DBI::dbConnect(
...
+
+The following types of funds were affected by this change:
+
+- Library funds
+- General assistance funds
+- Infrastructure funds (road and bridge)
+- Mental health and public health funds
+- Levy adjustments
+
+Most tax codes contain at least one of these types of agencies.
```
kyrasturgill, line 60: **[question, clarifying]:** Is that accurate? I tried to verify this but only find about 3,000 tax codes are connected to an agency that became a fund in 2024 - or did you mean PIN (since all of Chicago obviously is impacted by this).

vignettes/agencies.Rmd
```diff
@@ -47,74 +48,126 @@ ptaxsim_db_conn <- DBI::dbConnect(
...
-Before we query the relevant agency data, we first will need to check if any of the agencies of interest were impacted by the Clerk's 2024 updates to the data structure. We added new fields to the `agency_info` table to document these changes - we'll query this table and select only the agencies where the field `agency_change_24 = TRUE`.
+In 2024, the Clerk switched to reporting 78 agencies as funds underneath a separate agency. These agencies had always represented funds in the real world, but the Clerk reported them as independent taxing agencies prior to 2024. We need to account for this change when analyzing agencies and funds over time.
+
+The following types of funds were affected by this change:
+
+- Library funds
+- General assistance funds
+- Infrastructure funds (road and bridge)
+- Mental health and public health funds
+- Levy adjustments
```
kyrasturgill: **[nitpick, blocking]**: I'm not sure it makes sense to have levy adjustments listed here - I think all non home-rule agencies by default have this fund, so in past few years these agencies turned funds did have them, and we need to account for them, but they aren't necessarily standalone agencies that get turned to funds. Does that make sense?

vignettes/agencies.Rmd
```diff
@@ -47,74 +48,126 @@ ptaxsim_db_conn <- DBI::dbConnect(
...
+    agency_num_final, agency_name_final
+  )
+
+agency_crosswalk_info %>%
+  filter(str_detect(agency_name_prev, "CITY OF CHICAGO LIBRARY FUND")) %>%
+  datatable()
 ```
 
-Using the table above, we learn that the `agency_num` values for the agencies `CITY OF CHICAGO`, `CITY OF CHICAGO LIBRARY FUND`, `BOARD OF EDUCATION`, and `CHICAGO PARK DISTRICT` are `030210000`, `030210001`, `050200000`, and `044060000` respectively. We'll query all fields from the `agency` table and filter by their `agency_num`.
+Now let's see how to use this crosswalk to track Chicago agencies over time.
```
kyrasturgill, line 118: **[suggestion]**: Reading this through, I'm realizing the relationship between the City, CPKD and CPS may be a bit confusing especially since we're "folding in" the library which is a part of the City's levy, but the other two are separate. I know that this is a common point of confusion for people not as familiar with the City's agencies. What do you think about adding a footnote here (open to elsewhere) that explains this further - something like:

"CPS and Chicago Park District are technically sister agencies of the City of Chicago - meaning while they are closely tied to and collaborate with the City of Chicago, they are independent government bodies with their own taxing authority. Chicago Public Library is a department of the City of Chicago whose budget is proposed by the Mayor and approved by City Council."

vignettes/agencies.Rmd
kyrasturgill: Makes sense!

vignettes/agencies.Rmd
```diff
@@ -126,9 +179,9 @@ chi_agencies <- chi_agencies %>%
...
 
-[^2]: PTELL, as well as Chicago's property tax limitation ordinance, contain loopholes that allow taxing agencies to [increase their levies beyond the rate of inflation](https://civicfed.org/press/new-report-finds-illinois-property-tax-cap-law-not-working-intended). Let's see how Chicago's levy has fared compared to inflation since 2006.
+[^3]: PTELL, as well as Chicago's property tax limitation ordinance, contain loopholes that allow taxing agencies to [increase their levies beyond the rate of inflation](https://civicfed.org/press/new-report-finds-illinois-property-tax-cap-law-not-working-intended). Let's see how Chicago's levy has fared compared to inflation since 2006.
 
 To do so, we'll calculate the rate at which the levies have grown compared to the [CPI-U](https://www.bls.gov/cpi/). Fortunately CPI data, as reported by the Illinois Department of Revenue (IDOR) for purpose of PTELL calculatios, is available in the PTAXSIM data base in the `cpi` table.
```
kyrasturgill: Caught a typo mine:
```suggestion
To do so, we'll calculate the rate at which the levies have grown compared to the [CPI-U](https://www.bls.gov/cpi/). Fortunately CPI data, as reported by the Illinois Department of Revenue (IDOR) for purpose of PTELL calculations, is available in the PTAXSIM data base in the `cpi` table.
```

vignettes/agencies.Rmd
kyrasturgill: In the rendered version this text looks a bit odd being placed between the plot code drop down and the plot itself - do you mind moving it below the plot?

kyrasturgill: Nitpicking the way I phrased and formatted these dollar amounts.
```suggestion
Even though the City's levy has increased the most in percentage terms since 2006, it does not have the largest levy. The taxing agencies that rely most on property taxes are school districts and this is true for Chicago as well. The plot below shows the final tax extensions (meaning the final amount to be received by taxing agencies after the levies are validated by PTELL), for each of the three agencies. CPS is slightly more than double the City's extension in 2024 at about \$4 billion. Chicago Park District's extension is substantially less at around \$300 million.
```

DESCRIPTION
kyrasturgill: Good idea!

vignettes/agencies.Rmd
```diff
@@ -462,4 +513,113 @@ fund_type_table %>%
...
+  labs(
+    x = NULL,
+    y = "Final Levy (Millions)"
+  ) +
+  theme_minimal(base_size = 12)
+
+chi_library_pension_fund_plot
+```
+
+## Conclusion
```
kyrasturgill: I think this should be a section header to be consistent with introduction.
```suggestion
# Conclusion
```

vignettes/agencies.Rmd
```diff
@@ -462,4 +513,113 @@ fund_type_table %>%
...
+
+```{r}
+chi_library_pension_fund_plot <- chi_library_pension_fund %>%
+  ggplot(aes(x = as.integer(year), y = final_levy)) +
+  geom_line(linewidth = 0.5, color = "black") +
+  geom_point(color = "black") +
+  scale_x_continuous(n.breaks = 10) +
+  scale_y_continuous(
+    labels = scales::label_dollar(scale = 1e-6, suffix = "M"),
+    limits = c(5e6, NA)
```
kyrasturgill, line 615: **[suggestion, non-blocking]**: this is a simpler chart so not really necessary but for consistency we could put similar drop down for plot code - I'm open to pushback though!

NEWS.md
```diff
@@ -96,25 +100,19 @@ database and functions to handle these changes in the source data.
       vignette](https://ccao-data.github.io/ptaxsim/articles/tifs.html), which
       we have updated to include a TIF counterfactual with data for tax year
       2024.
```
kyrasturgill: Just noticing in the rendered changelog there seems an unnecessary line break at  `tax year 2024`

README.Rmd
```diff
@@ -267,7 +269,7 @@ The PTAXSIM backend database contains cleaned data from the Cook County Clerk, T
 ## Notes and caveats
```
kyrasturgill, line 269: I know this is beyond the scope of this PR, but what are your thoughts on adding a note here about how PINs with an EAV less than $150 will have a $0 bill but our `tax_bill()` function does not have that behavior built in?

"For PINs that have a final taxable EAV less than $150, the Cook County Treasurer will by default set the PIN's tax bill total to $0; note that we do not incorporate this behavior in the `tax_bill()` function which will lead our calculations to not match the final bill total."

Or something along those lines.

kyrasturgill (changes requested):
This is really awesome and I'm so impressed you incorporated all of the new data schema and vignette updates so quickly. I mainly just have a few clarifying questions and corrections of some typos I had made in the vignette. 
Let me know if there's anything you want to discuss further on Monday!

data-raw/agency/agency.R
jeancochrane: I _think_ that's right, though I sent a long Teams message to clarify, since it's confusing 😅 I updated the docstring in aedd2d59eb0779cd6420a39d8b0320c3c810eb77 to better communicate my understanding of the situation.

For the sake of posterity, here's my Teams message:

<details>
<summary>Expand to see full message</summary>

> [This point about levy adjustments](https://github.com/ccao-data/ptaxsim/pull/89/changes#r3175253584) is super helpful, thanks for the context! It lines up with what I see in the data, which is that each type of fund that used to get reported as an agency (general assistance, mental health, etc) had its own levy adjustment prior to 2024, but now all of those levy adjustments get folded into a single fund underneath the parent agency.
>
> I'm wondering if we should remove these funds from `agency_fund_crosswalk` altogether, or if we should include them with more documentation about the particulars of that case? It feels tricky because the agency number for the fund did change, so theoretically the crosswalk could be useful in a similar way as it is for the rest of the changing funds, but also the nature of the change is different than it is for the other funds. For example, general assistance funds and mental health funds continue to get reported as separate line items for an agency in 2024, so including them in the fund crosswalk makes intuitive sense; but the Clerk has rolled up all of the levy adjustments that were previously reported for each of those funds into one single levy adjustment for the agency, so it's actually a different type of change.
>
> Here's a motivating example that might help us decide what to do. If a user is analyzing levy adjustments to the mental health fund in Berwyn (formerly agency number 020020004), they would see levies like this between 2020-2023:
>
> ```R
> > agency_fund %>% filter(agency_num == '020020004', year >= 2020, fund_num == "408000") %>% select(year, agency_num, fund_num, levy)
> # A tibble: 3 × 4
> year  agency_num fund_num    levy
>  <chr> <chr>      <chr>    <int64>
> 1 2021  020020004  408000     13015
> 2 2022  020020004  408000     10229
> 3 2023  020020004  408000      3693
> ```
>
> However, following the 2024 fund change, all levy adjustments for all funds in Berwyn get rolled up into the parent agency (the township). The Berwyn township levy adjustment fund then looks like it has a big jump in 2024:
>
> ```R
> > agency_fund %>% filter(agency_num == '020020000', year == "2024", fund_num == "408000") %>% select(year, agency_num, fund_num, levy)
> # A tibble: 4 × 4
>  year  agency_num fund_num    levy
>  <chr> <chr>      <chr>    <int64>
> 1 2021  020020000  408000      8528
> 2 2022  020020000  408000      6774
> 3 2023  020020000  408000      2431
> 4 2024  020020000  408000     13299
>```
>
> The fund crosswalk instructs the user to fold these levy adjustments into the agency-level levy adjustment:
>
> ```R
> > agency_fund_crosswalk %>% filter(stringr::str_detect(agency_num, '^02002000'), fund_num == "408000")
> # A tibble: 3 × 5
>  year  agency_num agency_num_final fund_num fund_num_final
>  <chr> <chr>      <chr>            <chr>    <chr>         
> 1 2024  020020002  020020000        408000   408000        
> 2 2024  020020004  020020000        408000   408000        
> 3 2024  020020005  020020000        408000   408000     
> ```
>
> I think using the agency fund crosswalk as intended in this case -- i.e. joining to the fund levies to the crosswalk by `agency_num` and `fund_num`, then replacing them with `agency_num_final` and `fund_num_final` -- would "work" insofar as it would return a sensible value, the overall levy adjustment for the township over time. However, this value would be a different one than the one the user was looking for: Instead of continuing to return the mental health fund levy adjustment over time, it would return the township-level levy adjustment over time. That's the only value we can report on following the change to aggregate levy adjustments by township, but it's also not the same value the user was looking for. (The user would also end up with multiple rows with the same agency/fund number for a given year, and they would have to understand that they need to sum them to get the correct value for the township-level levy adjustment.)
>
> What do you think we should do in this case? My instinct is to continue including the levy adjustments in the fund crosswalk, but add a dedicated note about levy adjustments to the changelog and the agency vignette so that we can warn users that it might return an answer to a different question than the one they intended to ask. Still, I'm curious what you think as someone who has a lot more experience doing agency-level analysis than I do.
</details>

TL;DR: We agreed we would keep the levy adjustments in the crosswalk but add notes to the changelog and the agency vignette to explain how to use it for these specific funds.

data-raw/agency/agency.R
jeancochrane: Yup, that's right! I added a descriptive comment to both crosswalk blocks in f95232129240e7f8e2b96a5247a7028aa886a419 to make this clearer.

data-raw/agency/agency.R
```diff
@@ -576,59 +582,109 @@ agency_info <- agency_info %>%
...
-    agency_num,
-    agency_name,
-    agency_name_short,
-    agency_name_original,
-    major_type,
-    minor_type,
-    agency_num_24,
-    agency_name_24,
-    agency_change_24
+  inner_join(agency_crosswalk, by = "agency_num") %>%
```
jeancochrane, line 634: While making edits in response to your comments, I noticed that this left join + null filter is functionally identical to an inner join, so I changed it for the sake of simplicity in 0b662bf413b1a8eff91ea34ca54aa8030951a274. I also checked to confirm that the resulting crosswalk exactly matches the existing version in the DB:

```R
agency_fund_crosswalk_old <- DBI::dbGetQuery(
  ptaxsim_db_conn,
  "select * from agency_fund_crosswalk"
)
agency_fund_crosswalk_new <- agency_fund_crosswalk
agency_fund_crosswalk_old %>%
  full_join(
    agency_fund_crosswalk_new,
    by = c("agency_num", "fund_num"),
    suffix = c("_old", "_new")
  ) %>%
  filter(
    is.na(agency_num_final_old) |
      is.na(agency_num_final_new) |
      agency_num_final_old != agency_num_final_new |
      fund_num_final_old != fund_num_final_new |
      year_old != year_new
  )
```

vignettes/agencies.Rmd
jeancochrane: Here's some code that, if correct, shows that ~96% of tax codes in 2023 had at least one agency in the crosswalk, i.e. an agency that changed to a fund:

```R
library(dplyr)

ptaxsim_db_conn <- DBI::dbConnect(RSQLite::SQLite(), here::here("ptaxsim-2024.0.0-alpha.3.db"))

tax_code <- DBI::dbGetQuery(ptaxsim_db_conn, "select * from tax_code")
agency_crosswalk <- DBI::dbGetQuery(ptaxsim_db_conn, "select * from agency_crosswalk")

tax_code %>%
  filter(year == "2023") %>%
  left_join(
    agency_crosswalk %>%
      select(agency_num) %>%
      mutate(in_crosswalk = TRUE),
    by = c("agency_num")
  ) %>%
  group_by(tax_code_num) %>%
  summarise(
    has_crosswalk_agency = any(in_crosswalk, na.rm = TRUE),
    .groups = "drop"
  ) %>%
  summarise(prop_with_crosswalk_agency = mean(has_crosswalk_agency))
```

```
# A tibble: 1 × 1
  prop_with_crosswalk_agency
                       <dbl>
1                      0.962
```

Here are the counts:

```
# A tibble: 2 × 2
  has_crosswalk_agency     n
  <lgl>                <int>
1 FALSE                  169
2 TRUE                  4295
```

Does that match your understanding?

vignettes/agencies.Rmd
jeancochrane: Good point! I tweaked the language here and added a footnote in 2feba9979ee76029981b394d376a8c70103dfdf1.

vignettes/agencies.Rmd
jeancochrane: Good catch! For some reason GitHub wasn't rendering this suggestion correctly, so I applied it manually in 85b2f3e86f20b2a7bf2ea23420a23139899a2522.

vignettes/agencies.Rmd
jeancochrane: Good thinking, done in 6ea978dd6ed882a700db5d54c5b8298b0f8b0614!

NEWS.md
jeancochrane: Wow, good eye! For the curious: The root cause here was that Markdown interprets a line at this indent level starting with a number to indicate an ordered list. I fixed it in 10acb468a4e69107487a5996b03df92eb795c12b by pulling "2024" back onto the previous line.

README.Rmd
jeancochrane: We chatted on Teams and decided that we'll put up a quick PR to fix this bug instead!

data-raw/agency/agency.R
jeancochrane: I updated the changelog and the agency vignette to add some docs regarding levy adjustments in particular in a6a46364a3127470e6f9a28a3b5fb01ff0077168! Let me know if I got anything wrong, I don't fully understand recapture so I wouldn't be surprised if I misrepresented something.

vignettes/agencies.Rmd
jeancochrane: Good point! I removed this bullet in a6a46364a3127470e6f9a28a3b5fb01ff0077168, and added a separate section dedicated to explaining the nuance of working with levy adjustments.

vignettes/agencies.Rmd
jeancochrane: Agreed, done in 10c9b7ad523f2644d37d44c1c22582ad43637aa5.

data-raw/agency/agency.R
jeancochrane: This is a very helpful question! I doubt you will be the only person who is confused about what `year` means here. I decided to refactor this PR to rename `year` to `year_of_change` since otherwise I think it will be tempting for users to see the `year` column and assume they should include it in their join keys. That change is done in 43e99da7ad050f3db8670104e67040234ce3a599 and I'll plan to push a new export of the DB to `alpha-3` shortly.

jeancochrane (commented):
@kyrasturgill I believe I addressed all of your comments, though you should let me know if I missed anything! Mind taking one final look at this PR once tests pass?

data-raw/agency/agency.R
kyrasturgill: This looks great to me! Thanks!

data-raw/agency/agency.R
kyrasturgill: I really need to utilize `inner_join()` more

vignettes/agencies.Rmd
kyrasturgill: Yes this clearly is correct! I unfortunately accidentally deleted the scratch code where I did this calc, I wonder if I foolishly was measuring the inverse by mistake.

kyrasturgill:
This is good to go! Thanks for thoughtful responses to my questions.

kyrasturgill (approved):

PR #91, opened 2026-05-04
Add $0 bill output for PINs with an EAV less than $150

kyrasturgill:
This PR adds an exception to the `tax_bill()` function, defaulting the tax bill amount to zero when a PIN's final EAV is less than $150, mirroring the Treasurer which does not send tax bills to PINs with an EAV that is less than $150 and ultimately records the bill as $0. This ensures alignment between the `final_tax` output of the `tax_bill()` function and the `tax_bill_total` field which mirrors the final bill amount from the Treasurer's data.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -214,9 +214,9 @@ database and functions to handle these changes in the source data.
       PTAXSIM to analyze specific agencies over time so that you understand how
       to update your code.
 - **Corrected how the `tax_bill` function treats PINs with an EAV that is less
-  than $150, now defaulting the final tax output to zero. This aligns with the 
+  than $150, now defaulting the final tax output to zero**. This aligns with the 
   Treasurer's treatment of these PINs, so the tax bill calculation for these 
-  PINs will now match the what is shown on the final tax bill.**.
+  PINs will now match the what is shown on the final tax bill
   ([#91](https://github.com/ccao-data/ptaxsim/pull/91)).
     - **How this change affects you**: This improves the accuracy of the 
     `tax_bill` function to match what is displayed the Treasurer's tax bills.
--- a/NEWS.md
+++ b/NEWS.md
@@ -213,6 +213,14 @@ database and functions to handle these changes in the source data.
     - **How this change affects you**: You should read the vignette if you use
       PTAXSIM to analyze specific agencies over time so that you understand how
       to update your code.
+- **Corrected how the `tax_bill` function treats PINs with an EAV that is less
+  than $150, now defaulting the final tax output to zero. This aligns with the 
+  Treasurer's treatment of these PINs, so the tax bill calculation for these 
+  PINs will now match the what is shown on the final tax bill.**.
+  ([#91](https://github.com/ccao-data/ptaxsim/pull/91)).
+    - **How this change affects you**: This improves the accuracy of the 
+    `tax_bill` function to match what is displayed the Treasurer's tax bills.
+    This change has no impact on how to use the `tax_bill` function.
 
 ## Bug fixes
 
--- a/R/tax_bill.R
+++ b/R/tax_bill.R
@@ -242,6 +242,7 @@ tax_bill <- function(year_vec,
   dt[, tax_amt_pre_exe := round(eav * agency_tax_rate, 2)]
   dt[, tax_amt_post_exe := round(tax_amt_pre_exe - tax_amt_exe, 2)]
   dt[tax_amt_post_exe < 0, tax_amt_post_exe := 0]
+  dt[eav < 150, tax_amt_post_exe := 0]
   dt[, final_tax_to_tif := tax_amt_post_exe * tif_share]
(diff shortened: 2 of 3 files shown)


NEWS.md
```diff
@@ -213,6 +213,14 @@ database and functions to handle these changes in the source data.
     - **How this change affects you**: You should read the vignette if you use
       PTAXSIM to analyze specific agencies over time so that you understand how
       to update your code.
+- **Corrected how the `tax_bill` function treats PINs with an EAV that is less
+  than $150, now defaulting the final tax output to zero. This aligns with the 
+  Treasurer's treatment of these PINs, so the tax bill calculation for these 
+  PINs will now match the what is shown on the final tax bill.**.
+  ([#91](https://github.com/ccao-data/ptaxsim/pull/91)).
```
jeancochrane, line 220: **[Nitpick, optional]** Small tweak to align the bolding style with the rest of the doc (i.e. only the first sentence bolded) and clean up the PR link, which currently has some stray periods:


```suggestion
- **Corrected how the `tax_bill` function treats PINs with an EAV that is less
  than $150, now defaulting the final tax output to zero**. This aligns with the 
  Treasurer's treatment of these PINs, so the tax bill calculation for these 
  PINs will now match the what is shown on the final tax bill
  ([#91](https://github.com/ccao-data/ptaxsim/pull/91)).
```

R/tax_bill.R
```diff
@@ -242,6 +242,7 @@ tax_bill <- function(year_vec,
   dt[, tax_amt_pre_exe := round(eav * agency_tax_rate, 2)]
   dt[, tax_amt_post_exe := round(tax_amt_pre_exe - tax_amt_exe, 2)]
   dt[tax_amt_post_exe < 0, tax_amt_post_exe := 0]
+  dt[eav < 150, tax_amt_post_exe := 0]
```
jeancochrane: **[Question, non-blocking]** Something I should have thought about earlier: Do you know if the Treasurer's $150 limit applies to the pre- or post-exemption EAV? This code will apply it to the pre-exemption EAV, which may lead `tax_bill()` to miss some $0 bills if the Treasurer's logic actually depends on post-exemption EAV.

tests/testthat/test-tax_bill.R
```diff
@@ -412,4 +412,27 @@ test_that("Simplify FALSE / TRUE identical", {
...
+  WHERE av_board < 1000
+  AND av_board > 0
+  "
+    ) %>%
+    mutate(eav = av_board * eq_factor_final) %>%
+    filter(eav < 150) %>%
+    distinct(pin)
+
+  eav_lt_150_bills <- tax_bill(2006:2024, eav_lt_150_pins$pin) %>%
+    filter(eav < 150)
```
jeancochrane: **[Question, non-blocking]** I'm curious why this second filter is necessary?

jeancochrane (approved):
Thanks so much for the quick turnaround! A couple nits and questions below, but nothing serious.

R/tax_bill.R
kyrasturgill: AH I can't believe I didn't think of that! I believe it is post-exemption EAV. I did a test and found that there are PINs with pre-exemption EAV over $150 but a post-exemption under that have `tax_bill_total` equal to $0.

kyrasturgill: This is updated with commit [895fb0f](https://github.com/ccao-data/ptaxsim/pull/91/commits/895fb0f62f3d5bd458ddd62064d291609b9a5ac0)

tests/testthat/test-tax_bill.R
kyrasturgill: This is because the PINs examined here have had a taxable EAV less than 150 in at least one year, but not necessarily all years. If this filter is not included, the test fails because in some years those PINs had a non-zero bill.

jeancochrane: That makes sense, thanks for clarifying!

jeancochrane (approved):
Good to go!

kyrasturgill:
Documenting for posterity - I did some QC to ensure the PINs impacted by the change to the function were those we'd expect (taxable EAV < 150). In doing a quick calc of taxable EAV of `av_clerk * eq_factor_final - exe_total`, I found that farmland (class 239) can return "taxable EAV" > 150 (from the calc, that's not actually the farmland's taxable EAV as farmland is not equalized) but have a zero bill. This is because in the `tax_bill()` function we have:

 ` dt[class == "239", eav := av] # farmland is taxed on AV, not EAV`

So, this update for handling taxable eav < 150 will return $0 bill for farmland with av < 150. Turns out this rule aligns with what shows up in `tax_bill_total` - for all years, any 239 PIN with an AV < 150 has `tax_bill_total` equal to 0. So we're good, but wanted to just document in case anyone happens to inquire about how farmland is handled in this situation.

PR #93, opened 2026-05-05
Update DB version to 2024.0.0

jeancochrane:
This PR bumps the database version to 2024.0.0 and recreates it using all of the changes from the `2024-data-update` branch (plus #89, which this PR branches off of). This is the final step before 2024 data is ready for release.

The 2024 database object is now up on S3: `s3://ccao-data-public-us-east-1/ptaxsim/ptaxsim-2024.0.0.db.bz2`

The 2024 data is also now up on Athena. Proof:

```
select max(year) from tax.pin

> 2024
```

Tests are currently passing, but GitHub is having an incident and two of the R CMD Check workflows are sporadically failing as a result. I'll keep rerunning those till the issue gets resolved, but in the meantime, it shouldn't block review of this PR.

Connects https://github.com/ccao-data/ptaxsim/issues/59.


NEWS.md
```diff
@@ -230,11 +230,11 @@ database and functions to handle these changes in the source data.
...
       to update your code.
 - **Corrected how the `tax_bill` function treats PINs with an EAV that is less
-  than $150, now defaulting the final tax output to zero**. This aligns with the 
-  Treasurer's treatment of these PINs, so the tax bill calculation for these 
+  than $150, now defaulting the final tax output to zero**. This aligns with the
+  Treasurer's treatment of these PINs, so the tax bill calculation for these
   PINs will now match the what is shown on the final tax bill
   ([#91](https://github.com/ccao-data/ptaxsim/pull/91)).
-    - **How this change affects you**: This improves the accuracy of the 
+    - **How this change affects you**: This improves the accuracy of the
```
jeancochrane, line 237: I noticed there were some trailing spaces here while resolving merge conflicts, so I went ahead and cleaned them up.

vignettes/appeals.Rmd
```diff
@@ -384,8 +384,8 @@ sb_bills_all <- purrr::map_dfr(c("mailed", "certified", "board"), function(x) {
     setkey(year, tax_code, agency_num)
 
   tax_bill(
-    year = 2019,
-    pin = sb_pin_dt_stage$pin,
+    2019,
+    sb_pin_dt_stage$pin,
```
jeancochrane, line 388: I noticed these two issues in this vignette while testing `pkgdown::build_site()`. I'm not sure how we didn't notice this for so long, but `year` and `pin` are not actually the correct names for these arguments to `tax_bill()` (per [the docs](https://ccao-data.github.io/ptaxsim/reference/tax_bill.html#ref-usage), they should be `year_vec` and `pin_vec`) so rendering the vignette raises this error:

```R
Error in `tax_bill()`:
! argument 2 matches multiple formal arguments
Hide Traceback
    ▆
 1. ├─... %>% mutate(stage = x)
 2. └─dplyr::mutate(., stage = x)
 ```

Removing the incorrect `year`/`pin` argument names resolves the issue, and also aligns usage here with the way we call `tax_bill()` elsewhere in this vignette.

kyrasturgill: That is so strange -especially since I spent so much time in this vignette previously. I am wondering how this managed to render in the past?

kyrasturgill (approved):
Looks good to me - thanks for catching the weird error in the appeals vignette!

PR #97, opened 2026-05-11
Update testthat to Edition 3

### LINKED ISSUE #81: Upgrade to testthat 3rd edition
As we recently found in https://github.com/ccao-data/ccao/pull/51, testthat's `expect_known_hash()` assertion is starting to fail randomly across Ubuntu builds in our R packages, even though the functions under test have continued producing the same exact output on Ubuntu in every case I've debugged. The easiest fix that I have found so far has been to upgrade to [the 3rd edition of testthat](https://testthat.r-lib.org/articles/third-edition.html) so that we can switch to [`expect_snapshot_value()`](https://testthat.r-lib.org/reference/expect_snapshot_value.html), which is much more transparent and easier to debug when it fails. While it's possible to restrict 3rd edition usage to individual tests using the [`local_edition()`](https://testthat.r-lib.org/reference/local_edition.html?q=local_edition#null) config function, thereby keeping the rest of the tests on the legacy edition -- which is what we did in https://github.com/ccao-data/ccao/pull/48 -- my work on https://github.com/ccao-data/ccao/pull/51 has convinced me that it's worth updating all of our tests to 3rd edition, since there aren't that many breaking changes and the new edition includes a number of nice improvements alongside `expect_snapshot_value()`.

Given all this context, let's go ahead and upgrade this project to use the 3rd edition of testthat so that we can head off any `expect_known_hash()` failures in the future.

Damonamajor:
PR to upgrade testthat to a minimum version of 3 and follow-up to https://github.com/ccao-data/ccao/pull/51.

In unrelated notes, the tax rates vignette is randomly failing as seen [here](https://github.com/ccao-data/ptaxsim/actions/runs/25695680596/job/75443163338).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -57,7 +57,7 @@ Suggests:
     snakecase,
     stringr,
     tabulapdf,
-    testthat,
+    testthat (>= 3.0.0),
     tidyr,
     units,
     usethis
@@ -68,3 +68,4 @@ Remotes:
     ropensci/tabulapdf
 Config/Requires_DB_Version: 2024.0.0
 Config/Wants_DB_Version: 2024.0.0
+Config/testthat/edition: 3
--- a/tests/testthat/test-check.R
+++ b/tests/testthat/test-check.R
@@ -1,5 +1,3 @@
-context("test check_dt()")
-
 ##### TEST check_dt() #####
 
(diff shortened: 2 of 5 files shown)


jeancochrane (approved):
Thank you!

PR #102, opened 2026-07-07
2024 appeals vignette redo

kyrasturgill:
@ccao-jardine - this is ready for your comments! This version fully incorporates the method change in regards to ignoring the tentative equalizer. I still need to review the final rendered version to make sure that works but figured you could get a start on adding some of your comments.

@jeancochrane - no need to review yet!


vignettes/appeals.Rmd
```diff
@@ -52,167 +52,98 @@ ptaxsim_db_conn <- DBI::dbConnect(RSQLite::SQLite(), here("./ptaxsim.db"))
...
-  WHERE agency_name LIKE '%SCHAUMBURG%'"
+  WHERE agency_name LIKE '%CITY OF CHICAGO%'"
 )
 
-sb_agency_nums
+chi_agency_nums
 ```
 
-Here we can see the various taxing bodies around Schaumburg. The agency number we want is `031150000` for the `VILLAGE OF SCHAUMBURG`. With the agency number, we can find all of the *tax codes* that make up the municipality. To do so, we can again query PTAXSIM directly, this time looking in the `tax_code` table:
+Here we can see the various taxing bodies associated with the Chicago.  The agency number we want is `030210000` for the municipality, `CITY OF CHICAGO`. With the agency number, we can find all of the *tax codes* that make up the municipality. To do so, we can again query PTAXSIM directly, this time looking in the `tax_code` table:
```
ccao-jardine: Minor typo fix:

```suggestion
Here we can see the various taxing bodies associated with the City of Chicago.  The agency number we want is `030210000` for the municipality, `CITY OF CHICAGO`. With the agency number, we can find all of the *tax codes* that make up the municipality. To do so, we can again query PTAXSIM directly, this time looking in the `tax_code` table:
```

vignettes/appeals.Rmd
```diff
@@ -755,11 +724,11 @@ sb_cntr_plot <- bind_rows(
 <br>
 
 ```{r, out.width="100%", fig.height=8, echo=FALSE}
-sb_cntr_plot
+chi_cntr_plot
 ```
 
-Holding commercial appeals constant makes their effect on residential tax bills clearer: large commercial appeals shift the property tax burden back to residential property owners. In the case of Schaumburg, the drop in commercial AVs during the second level of review (at the Board of Review) contributed to a roughly 11% rise in the median residential tax bill.
+Holding commercial appeals constant makes their effect on residential tax bills clearer: large commercial appeals shift the property tax burden back to residential property owners. In the case of Chicago, the drop in commercial AVs during the second level of review (at the Board of Review) contributed to a roughly 9% rise in the median residential tax bill.
```
ccao-jardine: Vestigial typo from the original vignette?


```suggestion
Holding commercial assessments constant makes their effect on residential tax bills clearer: large commercial appeals shift the property tax burden back to residential property owners. In the case of Chicago, the drop in commercial AVs during the second level of review (at the Board of Review) contributed to a roughly 9% rise in the median residential tax bill.
```

vignettes/appeals.Rmd
```diff
@@ -52,167 +52,98 @@ ptaxsim_db_conn <- DBI::dbConnect(RSQLite::SQLite(), here("./ptaxsim.db"))
...
-    AND year = 2019
+    WHERE tax_code_num IN ({chi_tax_codes$tax_code_num*})
+    AND pin.year = 2024
     ",
     .con = ptaxsim_db_conn
   )
 )
-```
-
-## Mapping the area
```
ccao-jardine, line 107: I'll say it... I do miss the map! I appreciate that it takes forever to render because Chicago is so massive. I like your idea of rendering the map, saving it as a static image, and inserting that image along with commented code to reproduce it. Is it possible to render a map just to see what it looks like? Chicago is so large and heterogeneous that the map may be visually messy.

vignettes/appeals.Rmd
```diff
@@ -248,28 +179,32 @@ Finally, we can plot the index over time to see how reassessment and the subsequ
...
     color = "red"
   ) +
   geom_line(
-    data = sb_pins_summ,
-    aes(x = stage, y = idx, color = major_class, group = major_class),
+    data = chi_pins_summ,
+    aes(x = stage, y = perc_change, color = major_class, group = major_class),
     linewidth = 1.1
   ) +
   scale_color_brewer(name = "", palette = "Set1") +
```
ccao-jardine: For our color palette, can we use the CCAO color palette of the navy blue for residential, and the dark gold for commercial?

vignettes/appeals.Rmd
```diff
@@ -480,151 +429,168 @@ sb_bills_plot <- bind_rows(
...
+
+*A note about the equalizer*: Our appeals counterfactual experiment does not incorporate the impact of using the tentative equalizer rather than the final equalizer in calculating tax bills in the pre-appeal stage. Arguably the scenario in which pre-appeal values are used should use the tentative equalizer as we typically see a slight increase in the final equalizer once the Board's reductions are considered. The Illinois Department of Revenue (IDOR), who is responsible for calculating the equalizer, argues that the value of the equalizer does not impact tax bills. *However*, it does in fact impact the relative value of most homestead exemptions which are applied as flat reductions to a property's EAV. Because of the nature of most exemptions, impact of a higher equalizer does further shift the burden on to residential PINs. We do not incorporate the impact of the higher final equalizer here due to the challenges of rebuilding the taxing districts' tax base with the tentative equalizer, but we wanted to note it as a point of consideration - in a true and perfect counterfactual, we would likely see a slightly larger increase of residential tax bills in the `board` stage.
```
ccao-jardine: This is really great explanatory text! I worry that it's extremely helpful for the 4 people in Cook County who might benefit from this, but that it might introduce confusion for other readers. What do you think about having this text in the body of the PR, and/or having the text commented out?

ccao-jardine (changes requested):
This is looking really good, I'm excited for this update! A few suggested changes for your consideration, and let me know if we should rethink the map.

vignettes/appeals.Rmd
kyrasturgill: Still working on this!

ccao-jardine: If it's just impossible to render the map: because the content of this vignette is about commercial appeals, one idea could be to visualize Chicago's boundaries and only show a dot for the relevant commercial properties that received an appeal reduction. The saturation or darkness of the dot could scale with the assessment reduction, i.e., the difference between its 2024 mailed vs. final AV. Just a thought. Up to you!

ccao-jardine (approved):
Great, thanks for resolving my suggestions and for explaining your carefully thought-out change to the equalizer treatment! I fear I will never completely understand the equalizer. But your demonstration that using the final equalizer instead of the tentative equalizer provides more accurate estimates of total tax bills is convincing.

I might be looking at an old commit but it looks like there are a couple pre-commit checks to resolve before final merging. I know some of them, like the db connection, might waitto be corrected until the final merge.

I support whatever you decide with the map -- see comment for one more idea.

Nice work!

vignettes/appeals.Rmd
```diff
@@ -52,167 +52,98 @@ ptaxsim_db_conn <- DBI::dbConnect(RSQLite::SQLite(), here("./ptaxsim.db"))
...
 ```{r, echo=FALSE}
 # This is needed to build the vignette using GitHub Actions
-ptaxsim_db_conn <- DBI::dbConnect(
-  RSQLite::SQLite(),
-  Sys.getenv("PTAXSIM_DB_PATH")
-)
+#ptaxsim_db_conn <- DBI::dbConnect(
+ # RSQLite::SQLite(),
+  #Sys.getenv("PTAXSIM_DB_PATH")
+#)
```
jeancochrane: **[Nitpick, required]** Let's make sure to remove this commenting before we merge:

```suggestion
ptaxsim_db_conn <- DBI::dbConnect(
  RSQLite::SQLite(),
  Sys.getenv("PTAXSIM_DB_PATH")
)
```

vignettes/appeals.Rmd
```diff
@@ -230,15 +161,15 @@ sb_pins_summ <- sb_pins_all %>%
...
       paste0(year, "\n", stage),
       levels = c(
-        "2018\nboard", "2019\nmailed",
-        "2019\ncertified", "2019\nboard"
+        "2023\nboard", "2024\nmailed",
+        "2024\ncertified", "2024\nboard"
       )
     ),
-    idx = (med_av / med_av[stage == "2018\nboard"]) * 100
+    perc_change = (med_av - med_av[stage == "2023\nboard"])/med_av[stage == "2023\nboard"] *100
```
jeancochrane: **[Question, non-blocking]** Do we need to handle the possibility of $0 values in the `2023 board` stage, which would introduce a divide-by-zero error here?

vignettes/appeals.Rmd
```diff
@@ -293,115 +228,126 @@ sb_pins_av_plot <- ggplot() +
...
+      exe_freeze > 0,
+        eav - exe_total,
+        stage_eav - exe_total),
+      # We need to consider if a PIN is in a TIF to determine if any pre-appeal
+      # EAV will be added to the tax base or diverted to the TIF increment
+      eav_diff = case_when(
+        # Scenario 1: If both stage and actual EAV exceed the PIN's frozen EAV
+        # then the extra pre-appeal reduction EAV gets diverted to the TIF
+        taxable_eav > pin_frozen_eav &
+          pin_frozen_eav < taxable_stage_eav ~ 0,
```
jeancochrane: **[Nitpick, optional]** I would suggest we flip this particular conditional for the sake of readability. In the second condition in this `case_when` statement, we place the taxable EAV before the comparison operator when we compare taxable EAV to frozen EAV; that second condition is closely related to this one, so I think that it would make it easier for the reader to scan the difference between the two conditions if we stayed consistent in the order of the elements up for comparison.

```suggestion
          taxable_stage_av > pin_frozen_eav ~ 0, 
```

vignettes/appeals.Rmd
```diff
@@ -293,115 +228,126 @@ sb_pins_av_plot <- ggplot() +
...
+        taxable_eav > pin_frozen_eav &
+          pin_frozen_eav < taxable_stage_eav ~ 0, 
+        # Scenario 2: If the actual EAV is less than the frozen EAV and
+        # pre-appeal stage EAV is greater than frozen EAV, the diff between
+        # actual EAV and frozen EAV is added into the tax base
+          taxable_eav < pin_frozen_eav &
+          taxable_stage_eav > pin_frozen_eav ~ pin_frozen_eav - taxable_eav,
+        # Scenario 3: If the pre-appeal stage is less than the frozen EAV
+        # all of the EAV diff is added to the tax base.
+        pin_frozen_eav > taxable_stage_eav ~ taxable_stage_eav - taxable_eav,
```
jeancochrane, line 438: **[Question, non-blocking]** This condition depends on the assumption that post-appeal EAVs are always lower than pre-appeal EAVs, is that right? I wouldn't be surprised if that assumption always holds, I'm just curious if you think we need to handle the possibility that in some rare cases it might not.

vignettes/appeals.Rmd
```diff
@@ -293,115 +228,126 @@ sb_pins_av_plot <- ggplot() +
...
+        # For all of the non-TIF PINs, calculate the EAV diff
```
jeancochrane, line 439: **[Thought, non-blocking]** Note that we'll also drop into this conditional branch if `taxable_eav == pin_frozen_eav` and `taxable_stage_eav > pin_frozen_eav`, since the first two scenarios use exclusive comparison operators (`<` and `>`) when comparing `taxable_eav` to `pin_frozen_eav`. I don't think that's a problem, since in that specific case, `taxable_stage_eav - taxable_eav` is indeed the EAV difference that we want. I only note it because it confused me at first and made me wonder if perhaps we had a fifth scenario that we weren't properly handling.

vignettes/appeals.Rmd
```diff
@@ -293,115 +228,126 @@ sb_pins_av_plot <- ggplot() +
...
+        taxable_stage_eav - pin_frozen_eav,
+        0))
+  
+  # Calculate each TIF PIN's TIF share for the stages
+  tif_stage_share <- chi_pin_diff %>%
+    mutate(tif_pin_share = pin_stage_increment / taxable_stage_eav) %>%
+    select(pin, tif_pin_share)
+    
+  # Aggregate the EAV diff by tax code to which will enable joining to 
+  # tax district data
```
jeancochrane: **[Nitpick, optional]** Tiny typo (only flagging because this comment is part of code that will be rendered to the output doc):


```suggestion
  # Aggregate the EAV diff by tax code which will enable joining to 
  # tax district data
```

vignettes/appeals.Rmd
```diff
@@ -480,151 +428,165 @@ sb_bills_plot <- bind_rows(
...
 
 ```{r, warning=FALSE}
 # Again we need to calculate bills at each stage of appeal. This time it's more
 # complicated since we need to exclude commercial appeals
-sb_cntr_all <- purrr::map_dfr(c("mailed", "certified", "board"), function(x) {
-  sb_pin_dt_actual <- lookup_pin(2019, sb_pins$pin, stage = "clerk")
-  sb_pin_dt_mailed <- lookup_pin(
-    2019,
-    sb_pins$pin,
+chi_cntr_all <- purrr::map_dfr(c("mailed", "certified", "board"), function(x) {
```
jeancochrane, line 592: **[Thought, non-blocking]** I wonder if there's an easy way to share some of this code with the code that generates `chi_bills_all`. I suspect there's a reason Dan didn't refactor to share the code, but if we could do that, we could ensure that we're computing EAV diffs exactly the same in the three spots where we do that.

vignettes/appeals.Rmd
```diff
@@ -480,151 +428,165 @@ sb_bills_plot <- bind_rows(
...
-      ),
-    fill = TRUE
-  ) %>%
+      left_join(chi_pins) %>%
+    mutate(
+      tax_code = lookup_tax_code(2024, pin),
+      exe_total = rowSums(across(starts_with("exe_"))),
+      # We want to first calculate taxable EAV which is EAV minus exemptions.
+      # The stage taxable EAV should be the same as the actual taxable EAV
+      # if the PIN received a senior freeze exemption
```
jeancochrane: **[Question, non-blocking]** It seems like the old version of the code ignored exemptions here, possibly because these are commercial PINs. Do we want to do that here, too? Or do we include exemptions for the sake of insuring that the calculations are exactly the same in the three spots where we perform them?

jeancochrane (approved):
Nice work! Let me know if you need help resolving pre-commit failures.

vignettes/appeals.Rmd
kyrasturgill: I can see this being an issue if we're calculating percent change of all PINs and then taking median, but this is just the percent change of the medians none of which are $0.

vignettes/appeals.Rmd
kyrasturgill: Agreed!

vignettes/appeals.Rmd
kyrasturgill: I spent some time attempting this but ultimately gave up.. given that I didn't write the original code I had some difficulty finding a good and efficient way to refactor the code without blowing it up. I'm sure a better coder could easily do so, I just didn't feel like it was worth the effort it was going to take me.

vignettes/appeals.Rmd
kyrasturgill: I did delete these lines, given they are not relevant for commercial PINs.

vignettes/appeals.Rmd
jeancochrane: That's fine, thanks for trying!

vignettes/appeals.Rmd
kyrasturgill: Hmm good flag! I did a quick check and there are about 180 PINs in all of Chicago that do have an `av_board` greater than `av_certified` but many look like odd PINs (like the AV goes from $2 to $1). I am not inclined to build in an exception for those but lmk if you disagree!

jeancochrane: Sounds fine to me, thanks for looking into it!

vignettes/appeals.Rmd
```diff
@@ -60,159 +61,232 @@ ptaxsim_db_conn <- DBI::dbConnect(
...
+        "6": "Other",
+        "7": "Other",
+        "8": "Other",
+        "9": "Other"
+    }.get(first, "Other")
+
+color_map = {
+    "Residential": "#e07b7b",
+    "Commercial": "#7bafd4",
+    "Other": "#90c490"
```
ccao-jardine: For consistency with the rest of the vignette, can we modify the color palette to use our brand palette for navy, gold, and grey? 


```suggestion
color_map = {
    "Residential": "#29428D",
    "Commercial": "#FFAF00",
    "Other": "#787878"
```

vignettes/images/chi_parcel_class_map.png
ccao-jardine, line 1: Chicago's so beautiful! For consistency, can we use the same colors used in the vignette charts? I'm proposing Navy for residential, Gold for commercial, and Grey for others.

ccao-jardine (changes requested):
Great work! Just one requested color nitpick for the map png and the documented code -- goal is a consistent binding between class category and color throughout the vignette.

ccao-jardine (approved):
Excellent. Thank you!

